PR #28, opened 2021-02-09
Develop

khrulenko:


mgrinko (approved):

src/sortStudents.ts
```diff
@@ -3,6 +3,71 @@
...
+
+export function sortStudents(
+  studentsList: StudentType,
+  sortBy: SortField,
+  order: SortOrder
+) {
+  const studentsCopy = [...studentsList];
+  const sortOrder = (order === 'asc') ? [1, -1] : [-1, 1];
+
+  if (sortBy === 'name' || sortBy === 'surname') {
```
aionova, line 32: instead of multiple ifs it's better to use switch case construction

src/sortStudents.ts
```diff
@@ -3,6 +3,71 @@
...
+  AverageGrade = 'grades',
+}
+
+type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  studentsList: StudentType,
+  sortBy: SortField,
+  order: SortOrder
+) {
```
aionova, line 28: You can also try to make an interface for this function

aionova (approved):

PR #61, opened 2021-08-18
sort done

illiarerun:
I don't really know how i can fix this weird indentation in function params, i've tried a lot of ways but all of them failed((

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,7 +1,56 @@
-// describe Student type
-// create and export SortType enum
-// create SortOrder type
+type Student = {
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
+};
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,7 +1,56 @@
...
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+  const reducer = (sum: number, grade: number): number => sum + grade;
+  const gradesLengths = studentsCopy.map((student) => student.grades.length);
+  const maxGrades = Math.max(...gradesLengths);
+
+  studentsCopy.sort((a: Student, b: Student) => {
+    const aCopy: Student = order === 'asc' ? { ...a } : { ...b };
+    const bCopy: Student = order === 'asc' ? { ...b } : { ...a };
```
danheim, line 31: use enum instead of string

danheim (changes requested):

danheim (approved):

PR #63, opened 2021-08-19
Solution

alex-glotov67:


src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
 // describe Student type
 // create and export SortType enum
 // create SortOrder type
+export enum SortType {
```
varseniuk, line 4:
```suggestion
enum SortType {
```

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
+}
+
+enum SortOrder {
+  Asc = 'asc',
+  Desc = 'desc'
```
varseniuk, line 14:
```suggestion
  Ascending = 'asc',
  Descending = 'desc'
```

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+) :Student[] {
+  const studentsCopy: Student[] = [...students];
+
+  const sumOfGrades = (sum: number, grade: number) : number => sum + grade;
```
varseniuk, line 32: Just accept in this function the array as an argument and use `reduce` method here for this array. It will shorten your code

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+          return a[sortBy] - b[sortBy];
+        }
+
+        return b[sortBy] - a[sortBy];
+      });
+      break;
+
+    case SortType.AverageGrade:
+      studentsCopy.sort((a: Student, b: Student) => {
+        const averageA: number = a[sortBy].reduce(sumOfGrades, 0);
```
varseniuk, line 58: Change your `sumOfGrades` function and pass an array (`a[sortBy]` or `b[sortBy]`) to it

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+
+    case SortType.Married:
+      studentsCopy.sort((a: Student, b: Student) => {
+        if (a[sortBy] === b[sortBy]) {
+          return 0;
+        }
+
+        if (order === SortOrder.Asc) {
+          if (a[sortBy]) {
+            return 1;
```
varseniuk, line 77: You can use ternaries instead of ifs. It will make the code shorter

varseniuk (changes requested):
Good job, but let's make some changes

src/sortStudents.ts
alex-glotov67: Without export, tests didn't work. Its trow error about SortType was didn't export into the test file

varseniuk (approved):
Good job 🔥

Denchepornyuk (approved):

PR #69, opened 2021-08-21
solved

lera-yaskevych:


ihor-jpeg (approved):
your code looks great! actually a very nice and readable code. i'd way very good balance between wordiness and cognitive load! well done 🔥

VinnieJ-2k20 (changes requested):
Almost done, here is some advice on code improvement:
1. You can combine cases for Name + Surname and Age + Married using `student[propName]` prop access.
2. Currently your `AverageGrade` sort works using sum of all grades. If students will have different number of grades it may work incorrectly.

VinnieJ-2k20 (approved):
Good job 👍

PR #70, opened 2021-08-21
students sorted

ihor-jpeg:


src/sortStudents.ts
```diff
@@ -1,7 +1,74 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder = 'asc',
+): Student[] {
+  const copy = [...students].map((student) => ({
+    ...student,
+  }));
```
lera-yaskevych, line 26:
```suggestion
  const copy = [...students];
```
Would the code work the same if you simplify this, or am I missing something?

src/sortStudents.ts
```diff
@@ -1,7 +1,74 @@
...
+
+  type SortingCallback = (x: Student, y: Student) => number;
+
+  const makeSortingCallback = (
+    type: SortType,
+    sortOrder: SortOrder,
+  ): SortingCallback => {
+    const isAscendingOrder = (sortOrder) === 'asc';
```
lera-yaskevych, line 34: [
image
](url)

lera-yaskevych (approved):
Looks really good!

src/sortStudents.ts
ihor-jpeg: no, you are not missing anything. i have no idea why i created new objects. strange stuff on my side and nice catch on yours

src/sortStudents.ts
```diff
@@ -1,7 +1,74 @@
...
+
+    switch (type) {
+      case SortType.Surname:
+        return isAscendingOrder
+          ? (x, y): number => x.surname.localeCompare(y.surname)
+          : (x, y): number => y.surname.localeCompare(x.surname);
```
VinnieJ-2k20, line 40: This case and default can be combined using `object[propName]` prop access.

VinnieJ-2k20 (changes requested):
Need to update tests and sort by average grade
https://mateacademy.slack.com/archives/C0265RETQL9/p1629873475041600

PR #76, opened 2021-08-22
solution

DmytroMykoliv:


src/sortStudents.ts
```diff
@@ -2,6 +2,74 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyStudents: Student[] = students.map((student: Student) => (
+    ({ ...student })
+  ));
+
+  switch (sortBy) {
```
TarasHoliuk, line 32: It is a good decision to use swith construction here 👍

src/sortStudents.ts
```diff
@@ -2,6 +2,74 @@
...
+          b[sortBy].localeCompare(a[sortBy])
+        ));
+
+    case SortType.Age:
+      return order === 'asc'
+        ? copyStudents.sort((a: Student, b: Student) => a[sortBy] - b[sortBy])
+        : copyStudents.sort((a: Student, b: Student) => b[sortBy] - a[sortBy]);
+
+    case SortType.Married:
+      return copyStudents.sort((a: Student, b: Student) => {
```
TarasHoliuk, line 49: I'm not sure that it is a good idea, but you can sort boolean values using `+a[sortBy] - +b[sortBy]`

TarasHoliuk (approved):
Very well!

src/sortStudents.ts
DmytroMykoliv: But the Students.age type is a number in this case, and I'm not sure I should convert number to number. Maybe rewrite like this: a.age - b.age, for batter readable

DmytroMykoliv: Agree, it's a great idea ))

VinnieJ-2k20 (changes requested):
Need to update tests and sort by average grade correctly.
https://mateacademy.slack.com/archives/C0265RETQL9/p1629873475041600

VinnieJ-2k20 (approved):
Good job 👍

PR #80, opened 2021-08-23
Sort Students Task Done

sasha-morozov:


src/sortStudents.ts
```diff
@@ -2,6 +2,73 @@
...
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? sortedArr.sort((a, b) => (
+          (a.grades.reduce((sum, val) => sum + val, 0) / a.grades.length)
+          - (b.grades.reduce((total, v) => total + v, 0) / b.grades.length)
+        ))
+        : sortedArr.sort((a, b) => (
+          (b.grades.reduce((sum, val) => sum + val, 0) / b.grades.length)
+          - (a.grades.reduce((total, v) => total + v, 0) / a.grades.length)
+        ));
```
andrii-hryhoruk, line 66: You can create a callback function and use it for reduce, just to optimize Your code(just a suggestion)

Looks good and clean, nice choice to use ternary operator instead of if, else!

andrii-hryhoruk (approved):

src/sortStudents.ts
sasha-morozov: Thanks for the suggestion!)

Denchepornyuk (approved):

PR #104, opened 2021-08-26
first try

zadorojniiarsen:


Oleksandr-Zhitenko:
good job

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+
+export enum SortType {
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grade',
+}
+
+type SortOrder = 'asc' | 'desc';
```
varseniuk, line 17: use enum

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+      if (order === 'asc') {
```
varseniuk, line 34:
```suggestion
      if (order === 'asc') {
```
Use enum here instead of string

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+        studentsCopy.sort(
+          (x: Student, y: Student) => x.name.localeCompare(y.name),
+        );
+      } else {
+        studentsCopy.sort(
+          (x: Student, y: Student) => x.name.localeCompare(y.name),
```
varseniuk, line 40:
```suggestion
        studentsCopy.sort(
          (x: Student, y: Student) => x.name.localeCompare(y.name),
        );
      } else {
        studentsCopy.sort(
          (x: Student, y: Student) => x.name.localeCompare(y.name),
```

What the difference between these two sorting methods here? 🤔

src/sortStudents.ts
```diff
@@ -1,7 +1,96 @@
...
+        );
+      }
+
+      break;
+
+    case SortType.Surname:
```
varseniuk, line 46:
```suggestion
    case SortType.Name:
    case SortType.Surname:
```

You can combine these two cases because they are doing the same. Just compare by [sortBy]

varseniuk (changes requested):

varseniuk (approved):
Good job!

PR #105, opened 2021-08-26
task is done

Oleksandr-Zhitenko:


zadorojniiarsen (approved):

src/sortStudents.ts
```diff
@@ -2,6 +2,84 @@
...
+
+export enum SortType {
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+type SortOrder = 'asc' | 'desc';
```
varseniuk, line 21: Use enum

src/sortStudents.ts
```diff
@@ -2,6 +2,84 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      if (order === 'asc') {
```
varseniuk, line 37:
```suggestion
      if (order === 'asc') {
```
Use enum instead of string

varseniuk (changes requested):

varseniuk (approved):
👍

PR #107, opened 2021-08-27
Sort Students Solution

ihor-shapoval:


src/sortStudents.ts
```diff
@@ -1,7 +1,64 @@
...
+}
+export enum SortType {
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+type SortOrder = 'asc' | 'desc';
```
varseniuk, line 16:
```suggestion
type SortOrder = 'asc' | 'desc';
```
Use enum here

src/sortStudents.ts
```diff
@@ -1,7 +1,64 @@
...
+): Student[] {
+  const sortedStudents: Student[] = [...students];
+  const reducer = (acc: number, curr: number): number => {
+    return acc + curr;
+  };
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      if (order === 'asc') {
```
varseniuk, line 31:
```suggestion
      if (order === 'asc') {
```

Use enum instead of string

varseniuk (changes requested):
All is good, but let's use enum instead of strings

varseniuk (approved):
👍

PR #125, opened 2021-10-04
Solution

s-person27:


CouldSweeT (approved):

src/sortStudents.ts
```diff
@@ -1,7 +1,67 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyOfStudents: Student[] = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
```
aliceblunt8, line 26: I think your solution would be neater if you use switch inside the sort function itself, this way you change the compare function on different conditions

src/sortStudents.ts
```diff
@@ -1,7 +1,67 @@
...
+            - b[sortBy]
+              .reduce((acc: number, grade: number) => acc + grade)
+                / b[sortBy].length)
+        : copyOfStudents
+          .sort((a: Student, b: Student) => b[sortBy]
+            .reduce((acc: number, grade: number) => acc + grade)
+              / b[sortBy].length
+            - a[sortBy]
+              .reduce((acc: number, grade: number) => acc + grade)
+                / a[sortBy].length);
```
aliceblunt8, line 63: create function for calculating the average grade `calcAverageGrade`

src/sortStudents.ts
aliceblunt8: you can also write `[...students].sort()`

aliceblunt8 (changes requested):

PR #202, opened 2021-11-01
solution

Anton-Karpena:


olena1994 (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+  return grades.reduce((sum, num) => sum + num, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const newStudents: Student[] = students
+    .map((person: Student) => ({ ...person }));
```
VinnieJ-2k20, line 30: We don't mutate students` objects, so object copy is unnecessary here. Copy of array is enough.

VinnieJ-2k20 (approved):

PR #208, opened 2021-11-20
Create student sorter

dmtrhrytsak:


src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+
+  return totalGrades / grades.length;
+};
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsSample: Student[] = students.slice();
```
illiarerun, line 30: Come on, you can use spread operator here

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+      });
+      break;
+
+    case SortType.Surname:
+      studentsSample.sort((studentA, studentB) => {
+        return isAscending
+          ? studentA.surname.localeCompare(studentB.surname)
+          : studentB.surname.localeCompare(studentA.surname);
+      });
+      break;
```
illiarerun, line 48: You can actually combine these two cases as they are almost the same, I know you can do it)

illiarerun (changes requested):
Good job 🔥  Lets fix some thing now:

illiarerun:
And also, don't add unnecessary files to your commits, it'll be a big mistake at your future job

src/sortStudents.ts
dmtrhrytsak: Sure thing)) I was just lazy...

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
```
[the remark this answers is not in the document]
dmtrhrytsak: image

src/sortStudents.ts
dmtrhrytsak: But it's classic!😄

src/sortStudents.ts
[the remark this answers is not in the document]
illiarerun: Yeah, but union types can be confusing if they get bigger, and also enums are "named" type, so it's more convenient to use them in code. There's actually no "right" point of view on this subject, it depends more on you, but I think that using enums is better

illiarerun (approved):
GJ 🔥

PR #209, opened 2021-11-23
solution was added

JamesSunderlanddd:


dmtrhrytsak (approved):
I guess, it's more suitable to use switch for conditional checking rather than a bunch of ifs)

The first reason for that is it's more readable, the second your program will check all conditions even if the first one was true, not good for performance

But other than that, excellent job🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+    }
+  }
 
+  if (sortBy === SortType.AverageGrade) {
+    if (order === 'asc') {
+      studentsCopy.sort((s1, s2) => getAverageGrade(s1) - getAverageGrade(s2));
+    } else {
+      studentsCopy.sort((s1, s2) => getAverageGrade(s2) - getAverageGrade(s1));
+    }
+  }
```
illiarerun, line 62: Agree with Dima, use switch case construction here

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
-export enum SortType {
-  // describe SortType enum
+export type SortOrder = 'asc' | 'desc';
```
illiarerun, line 17: It'd be easier to use enum here

illiarerun (changes requested):
Good job 🔥  Let's work on code style now:

illiarerun (approved):
LGTM 🔥

PR #210, opened 2021-11-24
solution

Val-space:


src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+
+        if (name1 === name2) {
+          return 0;
+        }
+
+        if (order === 'asc') {
+          return name1 > name2 ? 1 : -1;
+        }
+
+        return name1 > name2 ? -1 : 1;
```
dmtrhrytsak, line 41: You could just compare it using `localeCompare`))

src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+
+        if (surname1 === surname2) {
+          return 0;
+        }
+
+        if (order === 'asc') {
+          return surname1 > surname2 ? 1 : -1;
+        }
+
+        return surname1 > surname2 ? -1 : 1;
```
dmtrhrytsak, line 59: Again `localeCompare `is a cure to all problems)

src/sortStudents.ts
dmtrhrytsak: Or even subtract one name from another, I believe it'll work

dmtrhrytsak (approved):
As always, some good stuff here) So good that it made the whole server to shut down! Good job✨

src/sortStudents.ts
Val-space: Why did I avoid it? It's so much easier with localeCompare

dmtrhrytsak: You just like making your life harder))

src/sortStudents.ts
```diff
@@ -1,16 +1,91 @@
...
+
+      break;
+
+    case SortType.Surname:
+      studentsDublicate.sort((student1, student2) => {
+        return order === 'asc'
+          ? student1.surname.localeCompare(student2.surname)
+          : student2.surname.localeCompare(student1.surname);
+      });
+      break;
```
illiarerun, line 48: Code in these cases is almost the same, you get rid of one of the code declarations by using one of `switch's` features

src/sortStudents.ts
```diff
@@ -1,16 +1,91 @@
...
+  grades:number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
```
illiarerun, line 15: If you add values to your enum you can use them in switch cases to work with needed properties without explicitly writing them

illiarerun (changes requested):
Good job 🔥  LEt's work on fixes now:

src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
 
 export interface Student {
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades:number[];
```
illiarerun, line 7:
```suggestion
  grades: number[];
```

illiarerun (approved):
LGTM 🔥

PR #213, opened 2021-11-25
Solution

dariia-romanova:


alex-kostomarov (approved):
Nice solution! Nothing to add)

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
+function getAvarageGrade(grades : number[]) : number {
+  const sum : number = grades.reduce((result, grade) => result + grade, 0);
```
ihor-jpeg, line 2:
```suggestion
function getAvarageGrade(grades: number[]): number {
  const sum: number = grades.reduce((result, grade) => result + grade, 0);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desk';
```
ihor-jpeg, line 23:
```suggestion
export type SortOrder = 'asc' | 'desс';
```

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+)
+  : Student[] {
```
ihor-jpeg, line 28:
```suggestion
export function sortStudents(
  students: Student[],
  sortBy: SortType,
  order: SortOrder,
): Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+      ));
+      break;
+
+    case (SortType.Surname):
+      sortedStudents.sort(({ surname: surname1 }, { surname: surname2 }) => (
+        order === 'asc'
+          ? surname1.localeCompare(surname2)
+          : surname2.localeCompare(surname1)
+      ));
+      break;
```
ihor-jpeg, line 46: is there a way to combine these two? you do the exact same actions. the only thing that's different is the name of the property

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+
+    case (SortType.Married):
+      sortedStudents.sort(({ married: married1 }, { married: married2 })
+      : number => (
+        order === 'asc'
+          ? +married1 - +married2
+          : +married2 - +married1
+      ));
+      break;
```
ihor-jpeg, line 63: be consistent with your formatting. sort by age case is correct

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+    case (SortType.Surname):
+      sortedStudents.sort(({ surname: surname1 }, { surname: surname2 }) => (
+        order === 'asc'
+          ? surname1.localeCompare(surname2)
+          : surname2.localeCompare(surname1)
+      ));
+      break;
+
+    case (SortType.Age):
+      sortedStudents.sort(({ age: age1 }, { age: age2 }) : number => (
```
ihor-jpeg, line 49:
```suggestion
      sortedStudents.sort(({ age: age1 }, { age: age2 }): number => (
```

ihor-jpeg (changes requested):
Good work! But I think it can be optimized even more. Let's also fix the formatting

src/sortStudents.ts
dariia-romanova: Do you mean that I move `:number` to a new line? I did so because of line length limitation

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      ));
+      break;
+
+    case (SortType.Married):
+      sortedStudents.sort(({ married: married1 }, { married: married2 })
+      : number => (
+        order === 'asc'
+          ? +married1 - +married2
+          : +married2 - +married1
+      ));
```
ihor-jpeg, line 54:
```suggestion
      sortedStudents.sort(
        ({ married: married1 }, { married: married2 }): number => (
          order === 'asc'
            ? +married1 - +married2
            : +married2 - +married1
        )
      );
```

ihor-jpeg (approved):
Well done!
Yeah, my prev comment was about that `:number` on a new line. a function's return type is related to the argument section, so it seems strange to see it on a separate line. I have left an example of how you can format it better.

PR #214, opened 2021-11-27
Solution

alex-kostomarov:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
+}
+
+function getAverageGrade(student: Student): number {
+  return (student.grades.reduce((a, b) => a + b) / student.grades.length);
 }
```
dariia-romanova, line 12: It's better to always use initial value for reduce method
Also, you could pass only grades array as an argument, instead of all object.

dariia-romanova (approved):
Good job)

src/sortStudents.ts
aionova: agree. if you need only grades, there's no need to pass the whole student object here

aionova (approved):
Good job overall! But I agree with Dariia's comments and you'd better fix them

PR #215, opened 2021-11-27
Solution

msaikun:


olya-shyian (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyOfStudents = [...students];
+  const isOrderedBy: boolean = order === SortOrder.increase;
```
ihor-jpeg, line 28: I think the name of this variable doesn't really say what it stands for. IMHO it would be better to name it `isAscending` or something like that

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+      break;
+
+    case
+      SortType.Age:
+      copyOfStudents.sort((prevStudent, nextStudent) => {
+        return isOrderedBy
+          ? prevStudent.age - nextStudent.age
+          : nextStudent.age - prevStudent.age;
+      });
+      break;
```
ihor-jpeg, line 47:
```suggestion
    case SortType.Age:
      copyOfStudents.sort((prevStudent, nextStudent) => {
        return isOrderedBy
          ? prevStudent.age - nextStudent.age
          : nextStudent.age - prevStudent.age;
      });
      break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+
+    case
+      SortType.AverageGrade:
+      copyOfStudents.sort((prevStudent, nextStudent) => {
+        const averageGrade = (student: Student): number => {
+          return student.grades.reduce(
+            (sum: number, grade: number) => sum + grade,
+          ) / student.grades.length;
+        };
```
ihor-jpeg, line 65: do not make your switch case unnecessarily complex. do not create methods inside it. rather create them outside and just use them here

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+
+        return isOrderedBy
+          ? averageGrade(prevStudent) - averageGrade(nextStudent)
+          : averageGrade(nextStudent) - averageGrade(prevStudent);
+      });
+      break;
 
+    default:
+      throw new Error('Error!');
```
ihor-jpeg, line 74: I am a user or a developer and I use your function. it tells me 'Error!'. I don't even understand what is going on. 
Error messages should be informative and let me know what error occurred so I can fix it.

ihor-jpeg (changes requested):
Great work! But let's work on the fixes

aionova (approved):

PR #219, opened 2021-11-29
Solution

olya-shyian:


msaikun (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+        };
+
+        return order === SortOrder.increase
+          ? averageGrade(prevStudent) - averageGrade(nextStudent)
+          : averageGrade(nextStudent) - averageGrade(prevStudent);
+      });
+      break;
 
+    default:
+      throw new Error('Error!');
```
ihor-jpeg, line 74: the idea behind throwing errors is to let the user or another developer know what exactly happened. it's not useful to just say that there was an error

ihor-jpeg (changes requested):
Well done, but let's fix a few couple things :)

ihor-jpeg (approved):
Noiiiice!

PR #220, opened 2021-11-29
Solution

vlad-tuzenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students:Student[],
+  sortBy:SortType,
+  order:SortOrder,
+):Student[] {
+  const copyOfArray:Student[] = students.map((student) => {
```
Vitaliy-0, line 25: not bad, in future you also can use [ ...students ] and taking average value when you need in future

Vitaliy-0 (approved):
GJ

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+    const avG = student.grades.reduce((a, b) => a + b) / student.grades.length;
+
+    return {
+      ...student,
+      averageGrade: avG,
+    };
+  });
```
ihor-jpeg, line 32: this is great. but there is a problem with such an approach: when I want to sort by Name, your code still performs map and reduce, even though sorting by name has nothing to do with getting the average grade. if your input becomes larger, you will experience slowdowns. do not perform unnecessary actions before you have to

ihor-jpeg (approved):
Good work! Give those comments a thought

PR #223, opened 2021-11-29
sort students

lera-permiakova:


Nadiyahr (approved):
GJ!

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+          return 0;
+        }
+
+        if (order === 'asc') {
+          return student1.married ? 1 : -1;
+        }
+
+        return student2.married ? 1 : -1;
+      });
+      break;
```
ihor-jpeg, line 61: you can combine these two. cast both of them to the number type and perform subtraction.

ihor-jpeg (approved):
Well done! 🔥  Have a look at the comments below

PR #225, opened 2021-11-29
Solution

4qwerty:


src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+        ));
+    }
+
+    case SortType.Surname: {
+      return newStudents.sort(isAscending
+        ? (student1: Student, student2: Student): number => (
+          student1.surname.localeCompare(student2.surname)
+        )
+        : (student1: Student, student2: Student): number => (
+          student2.surname.localeCompare(student1.surname)
```
kurtkobein, line 51:
```suggestion
        ? (student1, student2) => (
          student1.surname.localeCompare(student2.surname)
        )
        : (student1, student2) => (
          student2.surname.localeCompare(student1.surname)
```
You don't need to specify the type of callback function return value. (localeCompare always returns the number for strings).
Check code below

kurtkobein (approved):
Good job! :fire:

src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+  return grades.reduce((sum, num) => sum + num, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const newStudents: Student[] = students
+    .map((person: Student) => ({ ...person }));
```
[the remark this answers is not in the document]
ihor-jpeg: agreed. we only sort here. imagine we have the same group of students, but we just want them in a different order.

src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+          getAverageGrade(student1.grades) - getAverageGrade(student2.grades)
+        )
+        : (student1: Student, student2: Student): number => (
+          getAverageGrade(student2.grades) - getAverageGrade(student1.grades)
+        ));
+    }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default: throw new Error('Error');
```
ihor-jpeg, line 85: errors should be informative. if an error occurs I can see it without the message. the message should tell me what happened so I know how to fix it

src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+    }
+
+    case SortType.AverageGrade: {
+      return newStudents.sort(isAscending
+        ? (student1: Student, student2: Student): number => (
+          getAverageGrade(student1.grades) - getAverageGrade(student2.grades)
+        )
+        : (student1: Student, student2: Student): number => (
+          getAverageGrade(student2.grades) - getAverageGrade(student1.grades)
+        ));
```
ihor-jpeg, line 82: does it seem like you are writing the same code twice? it does to me. you can leave the part that is the same and put the condition on the part that is not the same. my point is you don't need a separate callback. all you need is to change the order of student1 and student2.

```suggestion
    case SortType.AverageGrade: {
      return newStudents.sort((student1: Student, student2: Student): number => (
          isAscending
            ? getAverageGrade(student1.grades) - getAverageGrade(student2.grades)
            : getAverageGrade(student2.grades) - getAverageGrade(student1.grades)
      ));
```

src/sortStudents.ts
```diff
@@ -1,16 +1,87 @@
...
+    }
+
+    case SortType.Age: {
+      return newStudents.sort(isAscending
+        ? (student1: Student, student2: Student): number => (
+          student1.age - student2.age
+        )
+        : (student1: Student, student2: Student): number => (
+          student2.age - student1.age
+        ));
```
ihor-jpeg, line 62: are you sure you need two different callbacks? can it be done with one?

ihor-jpeg (changes requested):
Good work! You have written quite a bit of unnecessary code. Let's make it look cleaner.

aionova (approved):

PR #227, opened 2021-11-29
solution

oleksiy-pochkayev:


ImIkator (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+export type SortOrder = 'asc'|'desc';
+
+function findAverage(grades:number[]):number {
```
ihor-jpeg, line 19:
```suggestion
function findAverage(grades: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+  return grades.reduce((prev, grade) => prev + grade) / grades.length;
 }
 
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students:Student[],
+  sortBy: SortType,
+  order:SortOrder,
+):object[] {
```
ihor-jpeg, line 27:
```suggestion
  students: Student[],
  sortBy: SortType,
  order: SortOrder,
): object[] {
```

let's keep formatting consistent

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+  const studentsCopy = [...students];
+
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
```
ihor-jpeg, line 32: it is generally not a good idea to use hardcoded values, such as strings or numbers. the reason is that it's hard to make sure it is consistent and it is hard to change because you will need to do so in many places. enum is made just for that. you can use your enum in many places and only change it in one place if you need it.
use your enum here

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+        return order === 'asc'
+          ? firstPersonAverageGrades - secondPersonAverageGrades
+          : secondPersonAverageGrades - firstPersonAverageGrades;
+      });
+
+      break;
+
+    default:
 
+      throw new Error('Unknown error');
```
ihor-jpeg, line 79: that's sorting. you can just return the original array at this point. throwing an error is okay too. but you need to tell the user or another developer what happened. an unknown error does not tell me anything. make the error message informative so I know why this error occurred

ihor-jpeg (changes requested):
Well done! Let's fix a few things below

aionova (approved):

PR #229, opened 2021-11-29
some message

dlugash:


src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
 
+function findAverage(arr:number[]):number {
+  return arr.reduce((sum, num) => (sum + num)) / arr.length;
+}
+
+export function sortStudents(
+  students:Student[],
+  sortBy:SortType,
+  order:SortOrder,
+  ):object[] | number {  // eslint-disable-line
```
DudarDaryna, line 29:
```suggestion
  ): Student[] | number {  // eslint-disable-line
```
I don't know why you wrote 'number' here, but about object - you can use your custom type

DudarDaryna (approved):

src/sortStudents.ts
DudarDaryna: Oh, I see why you use 'number' here :)

aionova (approved):

PR #231, opened 2021-11-29
Sorted students using TS

DudarDaryna:


src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+): Student[] {
+  const copyStudents = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyStudents.sort((prev, curr) => {
+        return order === 'asc'
+          ? prev[sortBy].localeCompare(curr[sortBy])
+          : curr[sortBy].localeCompare(prev[sortBy]);
```
SerhiiKirik, line 38: You can do like this
```suggestion
  const copyStudents = [...students];
  const check = order === 'asc';
  
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      copyStudents.sort((prev, curr) => {
        return check
          ? prev[sortBy].localeCompare(curr[sortBy])
          : curr[sortBy].localeCompare(prev[sortBy]);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+      copyStudents.sort((prev, curr) => {
+        return order === 'asc'
+          ? prev[sortBy] - curr[sortBy]
+          : curr[sortBy] - prev[sortBy];
+      });
+      break;
+
+    case SortType.Married:
+      copyStudents.sort((prev, curr) => {
+        return (order === 'asc')
```
SerhiiKirik, line 52: It is not necessary

```suggestion
        return order === 'asc'
```

SerhiiKirik (approved):
Well done 🚀

aionova (approved):
GJ!

PR #235, opened 2021-11-30
solution: ts_sort_students task

lanebx:


aldmpr (approved):
Good solution!

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+      );
+
+    case SortType.Surname:
+      return newArrayStudents.sort(
+        (a: Student, b: Student) => {
+          return asc
+            ? a.surname.localeCompare(b.surname)
+            : b.surname.localeCompare(a.surname);
+        },
+      );
```
ihor-jpeg, line 46: you can combine these two cases and write less code :)

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+
+    case SortType.Married:
+      return newArrayStudents.sort(
+        (a: Student, b: Student) => {
+          return asc
+            ? +a.married - +b.married
+            : +b.married - +a.married;
+        },
+      );
```
ihor-jpeg, line 64: same for these two

src/sortStudents.ts
```diff
@@ -1,16 +1,83 @@
...
+        (a: Student, b: Student) => {
+          const getAverageGrade = (student: Student): number => {
+            return (student.grades.reduce((acum, item) => acum + item, 0))
+            / student.grades.length;
+          };
+
+          return asc
+            ? getAverageGrade(a) - getAverageGrade(b)
+            : getAverageGrade(b) - getAverageGrade(a);
+        },
```
ihor-jpeg, line 77: the switch was made as a cleaner way to write if blocks. when you put a lot of stuff inside of each case it becomes difficult to understand what is happening. you can move outside all the logic that does not have to be inside a case. for example a method declaration. you only need to use it here. you can declare it outside of this block and make it less clunky

ihor-jpeg (changes requested):
Well done! Let's make your code more optimized and organized

aionova (approved):

PR #236, opened 2021-11-30
sortStudents solution

skachkov23:


aldmpr (approved):
GJ!

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+        (first, second) => {
+          return order === 'asc'
+            ? getAverageGrade(first[sortBy]) - getAverageGrade(second[sortBy])
+            : getAverageGrade(second[sortBy]) - getAverageGrade(first[sortBy]);
+        },
+      );
+      break;
 
+    default:
+      break;
```
ihor-jpeg, line 64: if you do not do anything in the default case, why use it?

ihor-jpeg (changes requested):
Good work! But let's do something about that default case. either make use of it or do not include it

ihor-jpeg (approved):
Well done!

PR #237, opened 2021-11-30
SOLUTION

aldmpr:


skachkov23 (approved):
GJ

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+export type SortOrder ='asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  sortedStudents.sort((previos: Student, current: Student) => {
```
ihor-jpeg, line 31:
```suggestion
  sortedStudents.sort((previous: Student, current: Student) => {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+        return current[sortBy] - previos[sortBy];
+
+      case SortType.Married:
+        if (previos[sortBy] && !current[sortBy]) {
+          return order === 'asc' ? 1 : -1;
+        }
+
+        if (!previos[sortBy] && current[sortBy]) {
+          return order === 'asc' ? -1 : 1;
+        }
```
ihor-jpeg, line 55: this is unnecessarily complex. you can use the same logic as in SortType.Age. you cannot subtract booleans from one another, but you can cast them to the number type and then do the subtraction

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+
+        return 0;
+
+      case SortType.AverageGrade:
+        if (order === 'asc') {
+          return getAverageGrade(previos.grades)
+          - getAverageGrade(current.grades);
```
ihor-jpeg, line 62:
```suggestion
          return getAverageGrade(previos.grades)
            - getAverageGrade(current.grades);
```

ihor-jpeg (changes requested):
Well done! But there are a few things to fix

aionova (approved):
GJ!

PR #243, opened 2021-11-30
Solution

alinaShchavinska:


tmrsl (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+  AverageGrade = 'grades',
+}
+
+export enum SortOrder {
+  desc = 'desc',
+  asc = 'asc',
+}
+
+function averageGrade(student: Student): number {
+  const totalSum = student.grades.reduce(
```
aionova, line 23: always add initial value in reduce

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+        aField = a[sortBy] as (number | string);
+        bField = b[sortBy] as (number | string);
+    }
+
+    let sortResult: number = 0;
+
+    if (typeof aField === 'string') {
+      sortResult = aField.localeCompare(bField as string);
+    } else {
+      sortResult = (aField as number) - (bField as number);
```
aionova, line 60: you'd better use keyof instead

aionova (changes requested):

src/sortStudents.ts
alinaShchavinska: Is it necessary? in this case, it works correctly🤞😉

aionova: sure, there are often many ways to do one thing in programming. I recommend you what I suppose to be the most appropriate here

aionova (approved):
GJ!

PR #246, opened 2021-11-30
Sort students

WorkAccCom:


aldmpr (approved):
Good

src/sortStudents.ts
```diff
@@ -1,16 +1,95 @@
...
+        sortedStudents.sort((studentA: Student, studentB: Student) => {
+          return avarageNum(studentA.grades) - avarageNum(studentB.grades);
+        });
+      } else {
+        sortedStudents.sort((studentA: Student, studentB: Student) => {
+          return avarageNum(studentB.grades) - avarageNum(studentA.grades);
+        });
+      }
+      break;
+    default: return {};
```
ihor-jpeg, line 91: this is very strange. your main function should not return just any object. it should return Student[]

src/sortStudents.ts
```diff
@@ -1,16 +1,95 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): object {
+  const sortedStudents: object = [...students];
```
ihor-jpeg, line 25: Student[]

ihor-jpeg (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,95 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): object {
```
ihor-jpeg, line 24: at the moment this function can return ANY object. but it to return ONLY Student[]

ihor-jpeg (changes requested):

ihor-jpeg (approved):
Well done!

PR #274, opened 2022-01-08
done

Volodymyr1806:


Rom911 (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+      return studentsWithAverageGrades.sort((a:Student, b: Student) => {
+        return Number(b.married) - Number(a.married);
+      });
+    case SortType.AverageGrade:
+      return studentsWithAverageGrades
+        .sort((a:StudentWithAverageGrade, b: StudentWithAverageGrade) => {
+          return b.averageGrade - a.averageGrade;
+        });
+    default: return studentsWithAverageGrades;
+  }
```
Olexandr6, line 84: Check 3 point in the [checklist](https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md)
You duplicate a lot bunch of code.

src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+
+export function sortStudents(students: Student[],
+  sortBy:SortType, order:SortOrder):StudentWithAverageGrade[] {
+  const studentsWithAverageGrades = students
+    .map((student: Student) => ({
+      ...student,
+      averageGrade: student.grades.reduce((prev, current) => {
+        return prev + current;
+      }, 0) / student.grades.length,
+    }));
```
Olexandr6, line 32: You don't need to add property `averageGrade` for each student, cause why do you need this prop if you just want to sort by `name, surname, age` etc. ? You don't need to store the prop always, it needed you only in case sorting by average grades.

Do you understand what I mean?

Olexandr6 (changes requested):
Check comments, your solution needs some optimization

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
-  // write your function
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? copyStudents.sort((a:Student, b: Student) => (
+          getAverage(a.grades) - getAverage(b.grades)
+        ))
+        : copyStudents.sort((a:Student, b: Student) => (
+          getAverage(b.grades) - getAverage(a.grades)
+        ));
+    default: return students;
```
Olexandr6, line 57:
```suggestion
        ));
        
    default: return students;
```

Olexandr6 (approved):
Looks good 🔥 Well done👍

PR #279, opened 2022-01-09
solution

andrii-koptiev:


Basil-Panasiuk (approved):
maybe better use switch case

andrii-koptiev:
> maybe better use switch case

have read few articles about using switch vs if and there no big difference between using them. switch could be more faster in some cases, but if is more readeable. Interesting to know mentor point. TY for review

danheim (approved):

PR #286, opened 2022-01-10
Solution

marina-tilniak:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+        case SortType.Age:
+          return student1.age - student2.age;
+        case SortType.Married:
+          if (student1.married === student2.married) {
+            return 0;
+          }
+
+          if (student1.married) {
+            return 1;
+          }
```
AlexAzalor, line 43: Це можна через звичайний сорт зробити

AlexAzalor (approved):

src/sortStudents.ts
marina-tilniak: можна, але тоді доведеться викликати sort 2 окремо для поля married. Тому, мені здається, краще передати туди callback

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+          return -1;
+        }
+
+        return 1;
+      case SortType.AverageGrade:
+        return averageGrade(student2.grades) - averageGrade(student1.grades);
+      default:
+        return 0;
+    }
+  });
```
Olexandr6, line 75: Check 3 point in the [checklist](https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md)

You duplicate a lot bunch of code.

Olexandr6 (changes requested):
Your solution works but needs some optimization. Avoid duplicating code!

src/sortStudents.ts
marina-tilniak: just i thought probably better use 2 switches and 1 if than 1 switch and 5 ifs

Olexandr6: @marina-tilniak you don't need to use `if`, you just can use ternary operator `?::`

marina-tilniak: But ternary operator is the same as if (or no?). And in general code with ternary operators looks better than with 2 switches, so I changed it, thanks.

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  function averageGrade(grades: number[]): number {
```
Olexandr6, line 26: Use verb in function naming
```suggestion
  function getAverageGrade(grades: number[]): number {
```

Olexandr6 (approved):
Well done 👍

PR #293, opened 2022-01-10
Develop

Sesile26:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
-
 export interface Student {
-  // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: [],
```
gnosis-frog, line 6: type number[] will be more specific for this property

gnosis-frog (approved):

src/sortStudents.ts
Olexandr6, line 6:
```suggestion
  grades: number[],
```

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #296, opened 2022-01-10
student sort

Toros-Zz:


src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+          - (studentA.grades.reduce(reduceCallback, 0)
+          / studentA.grades.length);
+        }
 
+        return (studentA.grades.reduce(reduceCallback, 0)
+                    / studentA.grades.length)
+          - (studentB.grades.reduce(reduceCallback, 0)
+          / studentB.grades.length);
+      default:
+        return 0;
```
YevheniiKorzun, line 72: Function sortStudents() should return an array, not a number.

YevheniiKorzun (changes requested):

YevheniiKorzun (approved):

Olexandr6 (changes requested):
Your solution works, but needs some optimization

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+
+  sortedStudents.sort((studentA, studentB) => {
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        if (order === 'desc') {
+          return studentB[sortBy].localeCompare(studentA[sortBy]);
+        }
+
+        return studentA[sortBy].localeCompare(studentB[sortBy]);
```
Olexandr6, line 39: It's better to use the ternary operator `?::` here. Try to retire your solution using the ternary operator

Olexandr6 (changes requested):
Almost done, check comments 👇

Olexandr6 (approved):
Looks good 🔥

PR #311, opened 2022-02-01
done!!!!

krasnobokydima:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  AverageGrade ='grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function average(arr: number[]): number {
+  return arr.reduce((acc: number, item: number) => acc + item, 0) / arr.length;
+}
```
Anna-Tkachenko, line 22: It's better to specify what this function do in function name (getAverageGrade, for example)

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const sorting = (a: Student, b: Student): number => {
+    let result: number;
+
+    switch (sortBy) {
+      case 'name':
+        result = comparing(a.name, b.name);
+        break;
+      case 'surname':
+        result = comparing(a.surname, b.surname);
+        break;
```
Anna-Tkachenko, line 46: Check point 3 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #316, opened 2022-02-08
Solution

DimaBrushnivskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+      return (order === 'asc')
+        ? copy.sort((a, b) => a.name.localeCompare(b.name))
+        : copy.sort((a, b) => b.name.localeCompare(a.name));
+
+    case SortType.Surname:
+      return (order === 'asc')
+        ? copy.sort((a, b) => a.surname.localeCompare(b.surname))
+        : copy.sort((a, b) => b.surname.localeCompare(a.surname));
```
aliceblunt8, line 40: combine these 2 cases

aliceblunt8: use sortBy variable to determine which values to sort
a[sortBy], for example

aliceblunt8 (changes requested):

PR #343, opened 2022-02-09
Solution

AndryKK:


jursylian (approved):
Perfect!

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copy = [...students];
+
+  const average = (student: Student): number => student.grades
```
Anna-Tkachenko, line 26: Function name should describe what this function do (getAverageGrade, for example)

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+    .reduce((acc, grade) => acc + grade) / student.grades.length;
+
+  function rev(a: number, b: number): number {
```
Anna-Tkachenko, line 29: What is rev?

Anna-Tkachenko (approved):
Good job
Just fix function names

src/sortStudents.ts
AndryKK: revers

PR #354, opened 2022-02-22
solution

SanyaBratashchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+        : b[sortby].localeCompare(a[sortby]);
+    }
+
+    if (typeof a[sortby] === 'boolean') {
+      return order === 'asc'
+        ? Number(a[sortby]) - Number(b[sortby])
+        : Number(b[sortby]) - Number(a[sortby]);
+    }
+
+    if (Array.isArray(a[sortby])) {
```
aspurgilus, line 41: I think it's better to use switch if "if's" more then 2.

aspurgilus (approved):
Good work. But I think there will be a lot of comments from mentor))

Ihor-Karpyn (approved):

PR #386, opened 2022-04-20
Students sort implemented using typescript

VETALEK:


src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
 }
```
yaremenko-maksym, line 17: You can use Name = 'name', so SortType.Name will return you 'name'.

src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+type CompareFunc = (student1?: Student, student2?: Student) => number;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+function getSorterBy(property: SortType, order: SortOrder = 'asc')
```
yaremenko-maksym, line 24: Use sort method

yaremenko-maksym (changes requested):
This code might use some optimization

src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
+      break;
+    }
+
+    case SortType.Surname: {
+      compareProperties = (student1, student2): number => {
+        return student1.surname.localeCompare(student2.surname);
+      };
+
+      break;
+    }
```
yaremenko-maksym, line 43:
``` suggestion
case SortType.Name:
case SortType.Surname: {
      compareProperties = (student1, student2): number => {
        return student[property].localeCompare(student[property]);
      };

      break;
    }
```

VETALEK:
But i still prefer `getSorterBy` implementation, so i don't need to check if order is `'asc`' or not in each `switch case`

yaremenko-maksym (commented):
All good!

MaksimusCRS (approved):

PR #389, opened 2022-04-21
solution

strukk311:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  const copyStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyStudents.sort((studentA, studentB) => {
+        return order === 'asc'
+          ? studentA[sortBy].localeCompare(studentB[sortBy])
+          : studentB[sortBy].localeCompare(studentA[sortBy]);
+      });
```
Anastasiia-Svintsova, line 41: You can write sort method once. And also you can check order once. For example
```
 copyStudents.sort((studentA, studentB) => {
    let a = studentA;
    let b = studentB;

    if (order === 'desc') {
      a = studentB;
      b = studentA;
    }

    switch (sortBy) {
      case SortType.Name:
      case SortType.Surname:...

      case SortType.Age:...
    }
  });

Anastasiia-Svintsova (changes requested):

MaksimusCRS (approved):

PR #400, opened 2022-04-22
sort was added

HelenPetrovska:


StasivPavlo (approved):
All good, but I don't understand why you need ```arrSort```

MaksimusCRS (approved):

PR #412, opened 2022-04-25
solution

barannikovArtem:


Jenia80 (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      a = secondEl;
+      b = firstEl;
+    }
+
+    switch (sortBy) {
+      case SortType.Name:
+        return a.name.localeCompare(b.name);
+
+      case SortType.Surname:
+        return a.surname.localeCompare(b.surname);
```
AlinaMatuschak, line 44: Can you convert elements on SortType to lowercase and use here sortBy?

```suggestion
      case SortType.Name:
      case SortType.Surname:
        return a[sortBy].localeCompare(b[sortBy]);
```

AlinaMatuschak (approved):

PR #418, opened 2022-04-26
done

Yurii837:


YegorKochetkov (approved):
LGFM :)

src/sortStudents.ts
```diff
@@ -1,16 +1,44 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    if (sortBy === SortType.Name || sortBy === SortType.Surname) {
+      return studentA[sortBy].localeCompare(studentB[sortBy]);
+    }
+
+    if (sortBy === SortType.AverageGrade) {
+      const avarageGradeA = studentA[sortBy].reduce((a, b) => a + b)
+      / studentA[sortBy].length;
```
AlinaMatuschak, line 35: Will be nice to create a separate function for that. Also, don't forget about the initial value for reduce

AlinaMatuschak (changes requested):
Nice, but switch case will be better here )

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  function averageGrad(arr: number[]): number {
+    return arr.reduce((a, b) => a + b) / arr.length;
```
AlinaMatuschak, line 26: How about an initial value for reduce?

AlinaMatuschak (approved):

PR #428, opened 2022-05-18
Develop

Maria-Lushchik:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
 
+export function sortStudents(students: Student[], sortBy: SortType, order: SortOrder): Student[] {
+  const arr: Student[] = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      if (order === 'asc') {
+        return arr.sort((a, b) => a.name.toLowerCase().localeCompare(b.surname.toLowerCase()));
```
aliceblunt8, line 26: use sortBy for the field name
then you can combine case SortType.Name and case SortType.Surname, cause they will be identical (sort strings)

```suggestion
        return arr.sort((a, b) => a[sortBy].toLowerCase().localeCompare(b[sortBy].toLowerCase()));
```

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      if (order === 'asc') {
+        return arr.sort((a, b) => +a.married - +b.married);
+      };
+
+      return arr.sort((a, b) => +b.married - +a.married);
+
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        return arr.sort((a, b) => {
+          const el1 = a.grades.reduce((total, current) => total + current) / a.grades.length;
```
aliceblunt8, line 55: create function getAverageGrades and use it everywhere

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #433, opened 2022-05-19
solution

vlad-demchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+        personBAverage = person2[sortBy]
+          .reduce((prev: number, item: number) => (prev + item)
+            , 0) / person2[sortBy].length;
+
+        return order === 'desc'
+          ? personBAverage - personAAverage
+          : personAAverage - personBAverage;
+
+      default:
```
alexandra-protyanova, line 58: you can return one of cases instead of 0

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+        return 0;
+    }
+  });
+
+  return copy;
```
alexandra-protyanova, line 63: and no need to return copy, toy are doing it in each case

alexandra-protyanova (approved):

src/sortStudents.ts
vlad-demchuk: i can't

src/sortStudents.ts
vlad-demchuk: i need

PR #434, opened 2022-05-19
Added solution to SortStudent task

dellachaise:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+          : firstStudent[sortBy].localeCompare(secondStudent[sortBy]);
+      case 'age':
+        return order === 'asc'
+          ? firstStudent[sortBy] - secondStudent[sortBy]
+          : secondStudent[sortBy] - firstStudent[sortBy];
+      case 'married':
+        return order === 'asc'
+          ? Number(firstStudent[sortBy]) - Number(secondStudent[sortBy])
+          : Number(secondStudent[sortBy]) - Number(firstStudent[sortBy]);
+      case 'grades':
```
alexandra-protyanova, line 44: you can use default for this case or another, instead of return 0)

alexandra-protyanova (approved):

src/sortStudents.ts
dellachaise: It's a great way to make code shorter, I agree with you! But in this case I am afraid that it wouldn't be so understandable. That's why I'll wait for Anya's comments)

Anna-Tkachenko (approved):
⭐️

PR #438, opened 2022-05-19
solution

lytvyshko:


src/sortStudents.ts
```diff
@@ -1,16 +1,47 @@
...
+function getAverageGrade(grades: number[]): number {
+  return grades.reduce((accum, grade) => accum + grade, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = students.slice();
```
AndriiRut, line 28: You can use `Spread operator` to make a copy of `objects` or `arrays`

AndriiRut (changes requested):
Good Job! (=

src/sortStudents.ts
```diff
@@ -1,16 +1,47 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
```
Anna-Tkachenko, line 30: Check point 1 in checklist

Anna-Tkachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,49 @@
...
+        ? studentsCopy.sort((a, b) => getAverageGrade(a[sortBy])
+        - getAverageGrade(b[sortBy]))
+        : studentsCopy.sort((a, b) => getAverageGrade(b[sortBy])
+        - getAverageGrade(a[sortBy]));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return order === 'asc'
+        ? studentsCopy.sort((a, b) => +a[sortBy] - +b[sortBy])
```
Anna-Tkachenko, line 46: Will be better to move all logic inside the sort method
And use default case for unexpected initial values or for throwing erros
This logic should be for 'born' and 'died' fields

Anna-Tkachenko (approved):

src/sortStudents.ts
lytvyshko: Have I understood you right?

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return studentsCopy.sort((a, b) => {
```
Anna-Tkachenko, line 33: Will be better to write all logic inside method sort
If you'll want to change variable name for students array or change some variables inside method sort you should do it 3 times in your code

Anna-Tkachenko (approved):

PR #440, opened 2022-05-20
Solution

59MrRobot:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(gradesArr: number[]): number {
+  return gradesArr.reduce((a, b) => a + b, 0) / gradesArr.length;
```
efdalyalcin, line 22: It's obvious you get the sum, but still I think you should use more clear names for the variables

efdalyalcin (approved):
Nice job!! 

Just a little comment: in line 22, I think you should use more clear variable names instead of a,b, although it is obvious you get the sum.

src/sortStudents.ts
59MrRobot: Okay noted, will do

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+          ? student1[sortBy] - student2[sortBy]
+          : student2[sortBy] - student1[sortBy];
+      });
+
+    case SortType.Married:
+      return sortedStudents.sort((student1: Student, student2: Student) => {
+        return order === 'asc'
+          ? +student1[sortBy] - +student2[sortBy]
+          : +student2[sortBy] - +student1[sortBy];
+      });
```
meforoma, line 54: cases `.Age` and `.Married` are much alike - try chaining two cases - it should work
```
case SortType.Age:
case SortType.Married:
 ...
```

meforoma (changes requested):

meforoma (approved):

PR #441, opened 2022-05-20
solution in one test run

AndriiRut:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
+// import { numberLiteralTypeAnnotation } from "@babel/types";
 
 export interface Student {
-  // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: [],
```
Andriyk0, line 8: write type of array value

Andriyk0 (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const array = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'name':
```
Anna-Tkachenko, line 34: Use SortType enum here

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+    case 'surname':
+      return array.sort((a: Student, b: Student) => (
+        (order === 'asc')
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy])
+      ));
+    case 'age':
+    case 'married':
+      return array.sort((a: Student, b: Student) => (
```
Anna-Tkachenko, line 43: You repeat this logic (calling method sort) 3 times, try to move all sort logic inside the method

Anna-Tkachenko (changes requested):

src/sortStudents.ts
AndriiRut: Should I place `switch` inside the sort method?
this variant will increase numbers of calls while sort be working, isn't it?

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  const array = [...students];
+
+  // switch (sortBy) {
+  //   case SortType.Name:
+  //   case SortType.Surname:
+  //     return array.sort((a: Student, b: Student) => (
+  //       (order === 'asc')
+  //         ? a[sortBy].localeCompare(b[sortBy])
+  //         : b[sortBy].localeCompare(a[sortBy])
+  //     ));
```
Denchepornyuk, line 39: remove comments

Denchepornyuk (approved):

PR #442, opened 2022-05-20
sort

Andriyk0:


yelnikov-andrii (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+export function sortStudents(
+  students:Student[],
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] {
+  const rezult:Student[] = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  function averGrade(person: Student): number {
```
Anna-Tkachenko, line 29:
```suggestion
  function getAverageGrade(person: Student): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students:Student[],
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] {
+  const rezult:Student[] = [...students];
```
Anna-Tkachenko, line 27: pay attention to variable naming, it's not rezult

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
-  // write your function
+  function averGrade(person: Student): number {
+    return person.grades.reduce((a, b) => a + b) / person.grades.length;
+  }
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? rezult.sort((people, people1) => people[sortBy].localeCompare(people1[sortBy]))
```
Anna-Tkachenko, line 37: Will be better to write all sort logic inside the sort method
If you'll want to change variable name for students array or some variables inside sort method you should do it 6 times in your code instead of 1

Anna-Tkachenko (approved):

PR #443, opened 2022-05-20
students

kitaminary:


src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
+  }
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      newStudents.sort((
+        person: Student,
+        person1: Student,
+      ) => {
+        if (order === 'asc') {
```
iwanttosmile4u, line 44: You can use ternary operator. In this case it'll be good, because you have simple checking.

iwanttosmile4u (commented):
Good!)

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
-export function sortStudents(students, sortBy, order) {
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
   // write your function
+  const newStudents: Student[] = [...students];
+
+  function avrGrades(arrayGrades: number[]): number {
```
Denchepornyuk, line 31: it's better to move this function outside sortStudents

Denchepornyuk (approved):

PR #444, opened 2022-05-20
solution

yelnikov-andrii:


Andriyk0 (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+      default:
+        copy.sort(
+          (secondStudent, firstStudent) => (
+            firstStudent.grades.reduce(
+              (a, b) => a + b, 0,
+            ) / firstStudent.grades.length)
+               - (secondStudent.grades.reduce(
+                 (a, b) => a + b, 0,
+               ) / secondStudent.grades.length),
+        );
```
Anna-Tkachenko, line 80: Don't repeat yourself: you wrote the same logic twice in your code just with changing places for some variables
Move all logic inside method sort and all checks also
Like this:
```
  students.sort((a, b) => {
     switch(...) {
       case SortType.name:
         if (order === 'asc) {
           // return something
         }

         // return something
     }
  })
```

Anna-Tkachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const copy = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  copy.sort((firstStudent: Student, secondStudent: Student) => {
+    switch (sortBy) {
+      case (SortType.Name):
+      case (SortType.Surname):
```
Denchepornyuk, line 29:
```suggestion
      case SortType.Name:
      case SortType.Surname:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+          return +firstStudent.married - +secondStudent.married;
+        }
+
+        return +secondStudent.married - +firstStudent.married;
+
+      default:
+        if (order === 'asc') {
+          return (firstStudent.grades.reduce(
+            (a, b) => a + b, 0,
+          ) / firstStudent.grades.length) - (secondStudent.grades.reduce(
```
Denchepornyuk, line 54: you use the same reduce twice, create a separate function for finding the average number

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #445, opened 2022-05-20
ts_sort_students

anastasiia-nikita:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+
+  switch (sortBy) {
+    case 'name':
+      if (order === 'asc') {
+        return newStudent.sort((a, b) => a.name.localeCompare(b.name));
+      }
+
+      return newStudent.sort((a, b) => b.name.localeCompare(a.name));
+
+    case 'surname':
```
iwanttosmile4u, line 40: You can combine a few cases. For example, you have the same logic for name and surname cases, so you can combine them: 
case 'name':
case 'surname': 
...and your logic

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      if (order === 'asc') {
+        return newStudent.sort((a, b) => a.surname.localeCompare(b.surname));
+      }
+
+      return newStudent.sort((a, b) => b.surname.localeCompare(a.surname));
+
+    case 'age':
+      if (order === 'asc') {
```
iwanttosmile4u, line 48: Also you can use ternary operator. I think, in this code it will be good, because you have simple checking.

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+}
+
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const newStudent = [...students];
```
Denchepornyuk, line 30: add type for `newStudent`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #446, opened 2022-05-20
Solution

iwanttosmile4u:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
-  // write your function
+  function getAverageGrade(person: Student): number {
+    const totalSum = person.grades.reduce((prev, curr) => (prev + curr), 0);
+
+    return totalSum / person.grades.length;
+  }
+
+  return copyOfStudentsArr.sort((student1, student2) => {
+    switch (sortBy) {
+      case SortType.Name:
```
alexandra-protyanova, line 35: in my opininon it's better to use 'name', 'surname' for `cases`, because your switch depends on values of `sortBy`, not  SortType

alexandra-protyanova (approved):
Well done!)

Anna-Tkachenko (approved):
⭐️

PR #447, opened 2022-05-20
Develop

khozhainov-aleksandr:


src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const newArray: Student[] = [...students];
+
+  if (order === 'asc') {
```
alexandra-protyanova, line 33: you can use ternary to check order, in the swith return 
like:
`case :
return  order === 'asc'
? option for true
: option for false`

alexandra-protyanova (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function sortAverageGrade(element: number[]): number {
```
Anna-Tkachenko, line 20: Pay attention to variables naming
```suggestion
function getAverageGrade(grades: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  let newArray: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      newArray = (order === 'asc' || order !== 'desc')
+        ? newArray.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
```
Anna-Tkachenko, line 37: Will be better to write all the logic inside the sort method
If you'll want to change students array variable name, or some variables inside the sort method you should do it 8 times in your code

Anna-Tkachenko (approved):

PR #448, opened 2022-05-20
Solution

RomekIvantsiv:


alexandra-protyanova (approved):
Well done 😎

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function avarageGrage(array: number[]): number {
```
Anna-Tkachenko, line 20:
```suggestion
function getAverageGrade(array: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+)
+  : Student[] {
+  const newStudents = [...students];
+
+  switch (sortBy) {
+    case 'name':
```
Anna-Tkachenko, line 33: use SortType enum here
```suggestion
    case SortType.Name:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+    case 'surname':
+      return order === 'asc'
+        ? newStudents
+          .sort((student1, student2) => student1[sortBy]
```
Anna-Tkachenko, line 37: Will be better to write all logic inside sort method
If you want to change the name of students array or variable names inside sort method you should do id 6 times in your code

Anna-Tkachenko (approved):

PR #449, opened 2022-05-21
solution

efdalyalcin:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      sum + current
+    ), 0);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return average / gradesAverage.length;
+}
+
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
```
59MrRobot, line 30: I think you should put each argument on its own line so it looks cleaner

59MrRobot (approved):
Good job, Efdal! 👍 

just one suggestion tho

src/sortStudents.ts
meforoma: upvoting @59MrRobot 's comment

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+          ? firstSt[sortBy] - secondSt[sortBy]
+          : secondSt[sortBy] - firstSt[sortBy];
+      });
+
+    case SortType.Married:
+      return sortedStudents.sort((firstSt: Student, secondSt: Student) => {
+        return order === 'asc'
+          ? Number(firstSt[sortBy]) - Number(secondSt[sortBy])
+          : Number(secondSt[sortBy]) - Number(firstSt[sortBy]);
+      });
```
meforoma, line 54: chain `.Age` and `.Married` - they're essentially the same
```
    case SortType.Age:
    case SortType.Married:
      ...
```

meforoma (changes requested):

meforoma (approved):

PR #454, opened 2022-05-22
solution

amitpatiljc:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+          ? +student1[sortBy] - +student2[sortBy]
+          : +student2[sortBy] - +student1[sortBy];
+      });
+
+    case SortType.AverageGrade:
+      return studentsCopy.sort((el1: Student, el2: Student) => {
+        const student1 = (el1.grades.reduce((sum, subject) => sum
+          + subject, 0)) / el1.grades.length;
+        const student2 = (el2.grades.reduce((sum, subject) => sum
+          + subject, 0)) / el2.grades.length;
```
LukaszGryziewicz, line 56: This reduce repeats here, maybe a separate function would make this more readable.

LukaszGryziewicz (approved):
Good job 👍  I left a little comment, please check it.

src/sortStudents.ts
amitpatiljc: yes mate! will do it 👍

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+          ? student1[sortBy] - student2[sortBy]
+          : student2[sortBy] - student1[sortBy];
+      });
+
+    case SortType.Married:
+      return studentsCopy.sort((student1: Student, student2: Student) => {
+        return order === 'asc'
+          ? +student1[sortBy] - +student2[sortBy]
+          : +student2[sortBy] - +student1[sortBy];
+      });
```
meforoma, line 49: cases `.Age` and `.Married` are effectively the same - combine then like you did with `.Name` and `.Surname`

src/sortStudents.ts
meforoma: yep, this makes total sense

meforoma (changes requested):

src/sortStudents.ts
amitpatiljc: Done

meforoma (approved):
👍 
although it would be even better if you don't repeat `.sort` for every case,
but use `Switch {case}` inside a `.sort` callback:
```
return studentsCopy.sort((student1: Student, student2: Student) => {
  switch (sortBy) {
    ...
  }
```

PR #456, opened 2022-05-22
sorte students

uliton:


src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const res = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  function averageGrades(arr: number[]): number {
```
vlad-demchuk, line 27: 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+    return arr.reduce((a: number, b: number) => a + b, 0) / arr.length;
+  }
+
+  switch (sortBy) {
+    case SortType.Name:
+      res.sort((st1, st2) => {
```
vlad-demchuk, line 33: it's better to use descriptive names of variables instead of st1, st2

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const res = [...students];
```
vlad-demchuk, line 25: it's better to use descriptive names of variables instead of res

vlad-demchuk (approved):
Good job)

src/sortStudents.ts
Denchepornyuk, line 27: it's better to move this function outside sortStudents

src/sortStudents.ts
Denchepornyuk: agree, use destructuring here and remove st1 and st2
also `res` isn't the best name for variable

Denchepornyuk (changes requested):

src/sortStudents.ts
Denchepornyuk, line 25: add type for `res`

src/sortStudents.ts
uliton: i don't understand what should i destructure here

Denchepornyuk (approved):

PR #461, opened 2022-05-22
solution

LukaszGryziewicz:


amitpatiljc (approved):
Good job 👍 Very clean code!

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+      );
+
+    case SortType.Married:
+      return studentsCopy.sort(
+        (student1: Student, student2: Student) => {
+          return order === 'asc'
+            ? Number(student1[sortBy]) - Number(student2[sortBy])
+            : Number(student2[sortBy]) - Number(student1[sortBy]);
+        },
+      );
```
meforoma, line 60: let's reduce code-lines count, - chain cases:
```
case SortType.Age:
case SortType.Married:
   ...
```
number will remain a number after `Number(number)`

meforoma (changes requested):

PR #462, opened 2022-05-23
Solution

JustDan17:


src/sortStudents.ts
```diff
@@ -1,16 +1,114 @@
...
+        default:
+          result = copiedStudents
+            .sort((obj1: Student, obj2: Student) => (
+              count(obj1[sortBy]) - count(obj2[sortBy])
+            ));
+      }
+      break;
+
+    default:
+      switch (true) {
```
vlad-demchuk, line 76: don't use two switch, one is enough

vlad-demchuk (changes requested):

JustDan17:
@Denchepornyuk Didn't change the boolean comment because I didn't convert anything to string there. And didn't change the last switvh comment'cause it's necessary for descending order.

src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copiedStudents: Student[] = [...students];
+  let result: Student[] = [];
+
+  switch (true) {
+    case order === 'asc':
+      switch (true) {
```
Denchepornyuk, line 38: don't write `true` in switch, you can use sortBy and write case without `===`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #467, opened 2022-05-23
Solution

rostyslav-pyliak:


src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+
+    case SortType.Age:
+      copyStudents.sort((student1: Student, student2: Student) => {
+        return order === 'asc'
+          ? student1[sortBy] - student2[sortBy]
+          : student2[sortBy] - student1[sortBy];
+      });
+      break;
+
+    case SortType.Married:
```
vlad-demchuk, line 50: you convert boolean to string, so you can combine the check with age

vlad-demchuk (approved):

src/sortStudents.ts
Denchepornyuk: agree, let's combine `case SortType.Age` and `case SortType.Married:` in one case

Denchepornyuk (changes requested):

src/sortStudents.ts
rostyslav-pyliak: Thanks. Fixed

PR #470, opened 2022-05-24
My solve

Yaroslav-Kononenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+): Student[] {
+  const afterSort = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  afterSort.sort((first, second) => {
+    let a: Student = first;
+    let b: Student = second;
+
+    function averageGrade(student: Student): number {
```
Denchepornyuk, line 31: it is better to move this function outside sortStudent function

Denchepornyuk (approved):

PR #479, opened 2022-06-01
Solution

naziamoff:


src/sortStudents.ts
```diff
@@ -1,16 +1,49 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const getAvg = (grades: number[]): number => grades
+    .reduce((a: number, b: number) => a + b) / grades.length;
```
aliceblunt8, line 24: imho better to move helper functions outside the main function

src/sortStudents.ts
```diff
@@ -1,16 +1,49 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
```
aliceblunt8, line 27: use enums here
```suggestion
    case SortType.Name:
    case SortType.Surname:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,49 @@
...
+      });
+    case 'married':
+    case 'age':
+      return [...students].sort((current, next) => {
+        return order === 'asc'
+          ? +current[sortBy] - +next[sortBy]
+          : +next[sortBy] - +current[sortBy];
+      });
+    case 'grades':
+      return [...students].sort((current, next) => {
```
aliceblunt8, line 41: also, you can do switch inside the sort itself (just another way to do this, this way you write `[...students].sort` only once)

aliceblunt8 (approved):
good job, as usual 🆒

PR #481, opened 2022-06-02
solution

nastia-sydorchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+  if (order === 'asc') {
+    switch (sortBy) {
+      case SortType.Name:
+        return [...students].sort((a: Student, b: Student) => {
+          return a.name.localeCompare(b.name);
+        });
+      case SortType.Surname:
+        return [...students].sort((a: Student, b: Student) => {
+          return a.surname.localeCompare(b.surname);
+        });
```
aliceblunt8, line 35: use sortBy variable 
```suggestion
      case SortType.Name:
      case SortType.Surname:
        return [...students].sort((a: Student, b: Student) => {
          return a[sortBy].localeCompare(b[sortBy]);
        });
```

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+        return [...students].sort((a: Student, b: Student) => {
+          switch (true) {
+            case a.married && b.married:
+              return 0;
+            case a.married:
+              return 1;
+            default:
+              return -1;
+          }
+        });
```
aliceblunt8, line 50: same here, also you can combine these two cases by converting booleans to Number

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+      case SortType.AverageGrade:
+        return [...students].sort((a: Student, b: Student) => {
+          const sumA: number
+          = a.grades.reduce((s: number, n: number) => s + n, 0);
+          const averageA = sumA / a.grades.length;
+          const sumB: number
+          = b.grades.reduce((s: number, n: number) => s + n, 0);
+          const averageB = sumB / b.grades.length;
```
aliceblunt8, line 58: create helper function getAverageGrade

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+          return averageA - averageB;
+        });
+      default:
+        return students;
+    }
+  } else {
```
aliceblunt8, line 65: remove this else, use ternary operator inside each case by order === 'asc' to determine the sort function

aliceblunt8 (changes requested):

src/sortStudents.ts
nastia-sydorchuk: Знімок екрана 2022-06-03 о 19 01 02
for whatever reason it does not work

nastia-sydorchuk:
Знімок екрана 2022-06-03 о 20 32 12
could you please also check why the status is not updated on the platform?

ArturErmolenko (approved):

PR #482, opened 2022-06-10
Sort students

YuriTaraday:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+)
+  : Student[] {
+  let result: Student[] = Object.assign([], students);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (sortBy === 'name' || sortBy === 'surname') {
```
Bohdan-Maksymiuk, line 28:
```suggestion
  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
```
and you can use it like here everywhere

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+    result = (order === 'asc')
+      ? result.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+      : result.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+  }
+
+  if (sortBy === 'grades') {
+    result.sort((a, b) => {
+      const firstLenght: number = a.grades.length;
+      const secondLenght: number = b.grades.length;
+      const first: number = a[sortBy].reduce((prev, el) => {
```
Bohdan-Maksymiuk, line 38: you may create separate helper function with logic of getting average grade and reuse it where you need

Bohdan-Maksymiuk (changes requested):
you may try to use `switch case` for `sortBy`

PR #484, opened 2022-06-17
solution

nazarroshchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const filteredStudents = [...students];
+
+  return filteredStudents.sort((a: Student, b: Student) => {
+    if (typeof a[sortBy] === 'string' && typeof b[sortBy] === 'string') {
```
aliceblunt8, line 54: use switch case and enum to identify sort technique by sortBy, also you can combine some cases when the data type is the same (like here with string)

```
switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+
+function sortArrTypeACS(a: number[], b: number[]): number {
+  return a.reduce((acc, el) => acc + el) / a.length
+  - b.reduce((acc, el) => acc + el) / b.length;
+}
+
+function sortArrTypeDESC(a: number[], b: number[]): number {
+  return b.reduce((acc, el) => acc + el) / b.length
+  - a.reduce((acc, el) => acc + el) / a.length;
+}
```
aliceblunt8, line 44: I think these helper functions are not needed

src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+
+function sortArrTypeACS(a: number[], b: number[]): number {
+  return a.reduce((acc, el) => acc + el) / a.length
+  - b.reduce((acc, el) => acc + el) / b.length;
+}
+
+function sortArrTypeDESC(a: number[], b: number[]): number {
+  return b.reduce((acc, el) => acc + el) / b.length
+  - a.reduce((acc, el) => acc + el) / a.length;
```
aliceblunt8, line 43: but you can create a helper fucntion getAverageGrade to find an average from a list of grades

aliceblunt8 (changes requested):

vkryvytskyy (approved):

PR #495, opened 2022-06-27
Solution

NikitaKarbazin:


src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] {
+  let newSortArrayWithStudents:Student[];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'Name':
```
Olexandr6, line 29: Let's use SortType here in cases check

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+          - nextStudent.grades
+            .reduce((prev, inn) => prev + inn) / nextStudent.grades.length;
+        })
+        : [...students].sort((prevStudent, nextStudent) => {
+          return nextStudent.grades
+            .reduce((prev, inn) => prev + inn) / nextStudent.grades.length
+            - prevStudent.grades
+              .reduce((prev, inn) => prev + inn) / prevStudent.grades.length;
+        });
+      break;
```
Olexandr6, line 73: Let's create a separate function to count average grade

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+  order: SortOrder,
+):Student[] {
+  let newSortArrayWithStudents:Student[];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'Name':
+      newSortArrayWithStudents = order === 'asc'
+        ? [...students]
```
Olexandr6, line 31: Let's make a copy of student array once not every time before `sort`

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+          .sort((prevStudent, nextStudent) => {
+            return Number(prevStudent.married) - Number(nextStudent.married);
+          })
+        : [...students]
+          .sort((prevStudent, nextStudent) => {
+            return Number(nextStudent.married) - Number(prevStudent.married);
+          });
+  }
```
Olexandr6, line 85: Let's better to create a separate case for married, don't put such logic in default case

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+
+    case 'Surname':
+      newSortArrayWithStudents = order === 'asc'
+        ? [...students]
+          .sort((prevStudent, nextStudent) => prevStudent.surname
+            .localeCompare(nextStudent.surname))
+        : [...students]
+          .sort((prevStudent, nextStudent) => nextStudent.surname
+            .localeCompare(prevStudent.surname));
+      break;
```
Olexandr6, line 47: Look, here you do the same in two cases with `Name` and `Surname`. Only one difference is you sort by a different prop. Let's combine these two cases into one.
For example:
```
case sortBy.Name:
case sortBy.Surname:
    students.sort(studnetA, studentB) => studnetA[sortBy].localCompare(studentB[sortBy])
```

Olexandr6 (changes requested):
Let's work on impovements👇

src/sortStudents.ts
NikitaKarbazin: image

Olexandr6 (approved):

PR #517, opened 2022-06-28
Solution

dimabelotskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+        sortedStudents.sort((a: Student, b: Student) => b[sortBy]
+          .localeCompare(a[sortBy]));
+      }
+      break;
+
+    case SortType.Age:
+    case SortType.Married:
+      if (order === 'asc') {
+        sortedStudents
+          .sort((a: Student, b: Student) => +a[sortBy] - +b[sortBy]);
```
Victance, line 46: it may be better to use Number to make the code more readable

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
 
+    default:
+      if (order === 'asc') {
+        sortedStudents.sort((a: Student, b: Student) => a.grades
+          .reduce(sum, 0) / a.grades.length
+            - b.grades.reduce(sum, 0) / b.grades.length);
+      } else {
+        sortedStudents.sort((a: Student, b: Student) => b.grades
+          .reduce(sum, 0) / b.grades.length
+            - a.grades.reduce(sum, 0) / a.grades.length);
```
Victance, line 61: I'd make a separate function for calculating an average grade

Victance (approved):
Good work🔥

src/sortStudents.ts
Olexandr6: Yeap, let's do this

src/sortStudents.ts
Olexandr6: Good advice

Olexandr6 (changes requested):

PR #534, opened 2022-06-29
Add sort sturents

ghost:


MariiaNH (approved):
Great job!

It's cool that you immediately combined similar cases into groups. The only thing I'm not sure about - is whether the else formulation is needed - but I leave this formatting to your discretion.

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function studentGrades(student: number[]): number {
```
Olexandr6, line 20:
```suggestion
function getAvergeGrade(student: number[]): number {
```

G-YaSha (approved):

PR #536, opened 2022-06-29
Sort Students done

garikr:


src/sortStudents.ts
```diff
@@ -1,16 +1,86 @@
...
+      }
+      break;
+
+    default:
+      break;
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return sortedStudents;
```
mykola-koval, line 85: you can return this value directly in cases

mykola-koval (approved):

PR #541, opened 2022-06-29
solution

LesyaSvk:


src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
+
+  return students
+    .sort((prev, next) => next.name.localeCompare(prev.name));
+}
+
+function sortBySurname(students: Student[], order: string): Student[] {
+  if (order === 'asc') {
+    return students.sort((prev, next) => prev.surname
+      .localeCompare(next.surname));
+  }
```
Victance, line 37: There is the same logic for sortByName and sortBySurname functions
It is better to merge them and use one function directly in the switch 

Also, you can take advantage of multi-case to simplify your code
image

src/sortStudents.ts
```diff
@@ -1,16 +1,108 @@
...
+
+function sortByMarried(
+  students: Student[],
+  order: string,
+): Student[] {
+  if (order === 'asc') {
+    return students.sort((prev, next) => {
+      return Number(prev.married) - Number(next.married);
+    });
+  }
```
Victance, line 73: same for sortByAge and sortByMarried

Victance (changes requested):
Nicely done👍

Let's DRY this code a little

PR #558, opened 2022-07-14
solution sort students

DimaShm:


valerii-derkach (changes requested):
to complicated, use switch case inside sort method, and use callback if it is necessary, for example for calculating average grades
https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md

valerii-derkach (approved):

PR #573, opened 2022-07-25
Added sortStudents function

LordAsmodey:


katerinahradinarova (approved):
👍

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+        }
+
+        return studentOne.name.localeCompare(studentTwo.name);
+
+      case SortType.Age:
+        if (order === 'asc') {
+          return studentOne.age - studentTwo.age;
+        }
+
+        return studentTwo.age - studentOne.age;
```
ivan-kadykalo, line 45: To avoid duplicating `if(order === 'asc')` in each case you can make `if(order === 'asc') { result.reverse()}` after you sort your array.

ivan-kadykalo (approved):

PR #581, opened 2022-07-26
Solution

LeonidPustovit:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+    case 'married':
+    case 'averageGrades':
+      if (order === 'asc') {
+        return studentsWithAverage.sort((student1, student2) => (
+          +student1[sortBy] - +student2[sortBy]
+        ));
+      }
 
+      return studentsWithAverage.sort((student1, student2) => (
+        +student2[sortBy] - +student1[sortBy]
```
DashaPoliazz, line 64:
```
order === 'asc' 
? studentsWithAverage
           .sort((student1, student2) => (+student1[sortBy] - +student2[sortBy]))
: studentsWithAverage
           .sort((student1, student2) => (+student2[sortBy] - +student1[sortBy]));
  
```

You can refactor your solution to form I described above. As for me it is hard to read nested if else conditions.

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+
+    obj.averageGrades = student.grades
+      .reduce((sum, grade) => sum + grade, 0) / student.grades.length;
+
+    return obj;
+  });
+
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
```
DashaPoliazz, line 43: Wow, I never met this before. Nice way to aggregate conditions.

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+  let studentsWithAverage: StudentWithAverage[] = [...students];
+
+  studentsWithAverage = studentsWithAverage.map((student) => {
+    const obj = { ...student };
+
+    obj.averageGrades = student.grades
+      .reduce((sum, grade) => sum + grade, 0) / student.grades.length;
+
+    return obj;
+  });
```
DashaPoliazz, line 39: I had the same idea to create 'averageGrades' field in copy. But I've been implemented little bit different solution. Now it seems to me that your solution it better.

DashaPoliazz (approved):
Nice solution.

src/sortStudents.ts
ivan-kadykalo, line 64: Use ternary operators

ivan-kadykalo (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): StudentWithAverage[] {
+  let studentsWithAverage: StudentWithAverage[] = [...students];
+
+  studentsWithAverage = studentsWithAverage.map((student) => {
```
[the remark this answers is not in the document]
LeonidPustovit: I am a bit confused by this `const studentsWithAverage: StudentWithAverage[] = ...studentsWithAverage.map(...`. We don't have students in this array, so This will not work, and If I try to use `const studentsWithAverage: StudentWithAverage[] = [...students].map(...`, then lint complains about averageGrades, which I want to add, because in Students we don't have such value. That is the reason I used that crutch.

ivan-kadykalo (approved):
👍💪

PR #582, opened 2022-07-26
Solution

Sergii2705:


src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+    },
+    [SortType.AverageGrade]: (): Student[] => {
+      return order === 'asc'
+        ? [...students].sort((a, b) => average(a.grades) - average(b.grades))
+        : [...students].sort((b, a) => average(a.grades) - average(b.grades));
+    },
+  };
+
+  return actions[sortBy]();
 }
```
illnino380, line 57: It's an interesting solution to make an object with every key is a sorting method! great)

illnino380 (approved):
Good solution!

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const average: (arr: number[]) => number
+    = (arr) => arr.reduce((sum, el) => sum + el) / arr.length;
```
ivan-kadykalo, line 26: Replace helping function outside of main function

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const actions: {[name: string]: () => Student[]} = {
+    [SortType.Name]: (): Student[] => {
```
ivan-kadykalo, line 29: Interesting solution, but it is hard to read + you make a lot of copies and sort for every case even when you need to do it only once. 
Use switch case construction and combine similar cases

ivan-kadykalo (changes requested):

src/sortStudents.ts
Sergii2705: I agree with you that there is a lot of copy-paste, but I do not agree that "it is hard to read". It's subjective, it's easier for me to read.

ivan-kadykalo (approved):

PR #585, opened 2022-07-26
solution

DmitriyTeteryn:


AleksandrYuschenko (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const avarage = (numbers: number[]): number => (
+    numbers.reduce((acc, num) => acc + num, 0) / numbers.length
+  );
```
ivan-kadykalo, line 26: Will be better to replace this function outside of sortStudents to avoid creating it on each call

ivan-kadykalo (changes requested):

src/sortStudents.ts
DmitriyTeteryn: fix it

ivan-kadykalo (approved):
GJ💣

PR #590, opened 2022-07-26
Done

Lentyaeev:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+  const averageGrade = (grades: number[]): number => {
+    return grades
+      .reduce((sum: number, value: number) => sum + value, 0) / grades.length;
+  };
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const copy = [...students];
+
+  if (sortBy === 'grades') {
```
blcts, line 32: [CODE STYLE] - If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks :(

blcts (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  };
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const copy = [...students];
+
+  switch (sortBy) {
+    case 'grades': copy.sort((a: Student, b: Student) => {
+      return order === 'asc' ? averageGrade(a.grades) - averageGrade(b.grades)
+        : averageGrade(b.grades) - averageGrade(a.grades);
```
ivan-kadykalo, line 35: Hard reading code. Make each condition on new line. 
```suggestion
      return order === 'asc' 
        ? averageGrade(a.grades) - averageGrade(b.grades)
        : averageGrade(b.grades) - averageGrade(a.grades);
```

ivan-kadykalo (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const copy = [...students];
+
+  switch (sortBy) {
+    case 'grades': copy.sort((a: Student, b: Student) => {
+      return order === 'asc'
+        ? averageGrade(a.grades) - averageGrade(b.grades)
+        : averageGrade(b.grades) - averageGrade(a.grades);
+    }); break;
```
ivan-kadykalo, line 37: if you have return you don't need to use break
```suggestion
    });
```

ivan-kadykalo (approved):
Fix my comment

PR #591, opened 2022-07-26
solution

Vova-Balyshev:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  AverageGrade
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+/* eslint-disable max-len */
+export function sortStudents(students: Student[], sortBy: SortType, order: SortOrder): object[] {
+  const studentsCopy = students.map((student) => ({ ...student }));
```
rodiontseva, line 22: maybe would be better like this:

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+    return studentsCopy
+      .sort((a, b) => (a.grades.reduce((x, y) => x + y) / a.grades.length) - (b.grades.reduce((x, y) => x + y) / b.grades.length));
+  }
+
+  if (sortBy === SortType.AverageGrade && order === 'desc') {
+    return studentsCopy
+      .sort((a, b) => (b.grades.reduce((x, y) => x + y) / b.grades.length) - (a.grades.reduce((x, y) => x + y) / a.grades.length));
+  }
+
+  return studentsCopy;
```
rodiontseva, line 66: https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md 
[CODE STYLE] - If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks

rodiontseva (changes requested):
good job! but it would be better to use the switch.

src/sortStudents.ts
Vova-Balyshev: Не заглянув в чекліст((

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+        return order === 'asc'
+          ? a.name.localeCompare(b.name)
+          : b.name.localeCompare(a.name);
+
+      case SortType.Surname:
+        return order === 'asc'
+          ? a.surname.localeCompare(b.surname)
+          : b.surname.localeCompare(a.surname);
```
ivan-kadykalo, line 41: Combine similar logic in one case
Use computed keys
For example
`a[sortBy].localeCompare(b[sortBy])`

ivan-kadykalo (changes requested):

src/sortStudents.ts
Vova-Balyshev: @ivan-kadykalo я не зовсім зрозумів помилку, якщо я міняю на a[name].localeCompare(b[name]), то в мене цілий ряд помилок вилазить :
'name' is deprecated.ts(6385)
lib.dom.d.ts(17693, 5): The declaration was marked as deprecated here.
Unexpected use of 'name'.eslint[no-restricted-globals](https://eslint.org/docs/rules/no-restricted-globals)
'name' is not defined.eslint[no-undef](https://eslint.org/docs/rules/no-undef)
Type 'void' cannot be used as an index type.ts(2538)

ivan-kadykalo (approved):

PR #595, opened 2022-07-26
Solution

blcts:


src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+      person.sort((a, b) => a.name.localeCompare(b.name));
+
+      break;
+
+    case SortType.Surname:
+      person.sort((a, b) => a.surname.localeCompare(b.surname));
+
+      break;
```
ivan-kadykalo, line 40: Combine similar logic to one case and use computed keys
```suggestion
    case SortType.Name:
    case SortType.Surname:
      person.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));

      break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+
+    case SortType.Age:
+      person.sort((a, b) => b.age - a.age).reverse();
+
+      break;
+
+    case SortType.Married:
+      person.sort((a, b) => +b.married - +a.married).reverse();
+
+      break;
```
ivan-kadykalo, line 50: And here

src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+
+    case SortType.AverageGrade:
+      person.sort((a, b) => (
+        a.grades.reduce((firstVal, secondVal) => firstVal + secondVal, 0)
+        / a.grades.length
+        - b.grades.reduce((firstVal, secondVal) => firstVal + secondVal, 0)
+        / b.grades.length
```
ivan-kadykalo, line 57: Make helping function outside of your main function and use here

src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+    default:
+      break;
+  }
+
+  if (order === 'desc' && sortBy === SortType.AverageGrade) {
+    person.sort((a, b) => (
+      b.grades.reduce((firstVal, secondVal) => firstVal + secondVal, 0)
+      / b.grades.length
+      - a.grades.reduce((firstVal, secondVal) => firstVal + secondVal, 0)
+      / a.grades.length
```
ivan-kadykalo, line 71: and here

ivan-kadykalo (changes requested):
And don't forget about additional reverse🙂
Good job🤘

src/sortStudents.ts
blcts: I don't think that here is a good idea, cause I need to add Number(), and its unreadable Number(number).I mean why do this :)

ivan-kadykalo (approved):
🚀

PR #597, opened 2022-07-26
solution

vitmixG:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(students:Student[], sortBy: SortType,
+  order: SortOrder): Student[] {
```
Oleksandr-Pustovit, line 20: You should choose the strategy of how to write parameters inline or from the new line and use it for EVERY parameter
E-g: sortStudents(students:Student[], sortBy: SortType, order: SortOrder) or 
```suggestion
export function sortStudents(
  students: Student[],
  sortBy: SortType,
  order: SortOrder
): Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  const res = [...students];
```
Oleksandr-Pustovit, line 21: _comment_
I bet there is a better naming for this variable

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  res.sort((firstEl, secondEl) => {
+    const orderElement = {
+      first: firstEl,
+      second: secondEl,
+    };
+
+    if (order === 'desc') {
+      orderElement.first = secondEl;
+      orderElement.second = firstEl;
+    }
```
Oleksandr-Pustovit, line 35: _comment_
you can use ternary operator here. But it is just a comment, can't see any real benefits exept less code.
```suggestion
    const orderElement = {
      first: order === 'asc' ? firstEl : secondEl,
      second: order === 'asc' ? secondEl: firstEl,
    };
```

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+
+      case SortType.Age:
+        return orderElement.first[sortBy] - orderElement.second[sortBy];
+
+      case SortType.Married:
+        return (+orderElement.first[sortBy]) - (+orderElement.second[sortBy]);
 
+      default:
+        return getAverage(orderElement.first[sortBy])
+        - getAverage(orderElement.second[sortBy]);
```
Oleksandr-Pustovit, line 51: _comment_
`default` is usually used to throw errors if no cases could be used. May be it is a good idea to add a separate `case` for averageGrades.

Oleksandr-Pustovit (approved):
Great!
There are several comments but it is mostly a different point of view.

src/sortStudents.ts
vitmixG: Thank you for useful comments🔥😉

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return orderElement.first[sortBy]
+          .localeCompare(orderElement.second[sortBy]);
+
+      case SortType.Age:
+        return orderElement.first[sortBy] - orderElement.second[sortBy];
+
+      case SortType.Married:
+        return (+orderElement.first[sortBy]) - (+orderElement.second[sortBy]);
```
ivan-kadykalo, line 45: You can combine this 2 cases

ivan-kadykalo (approved):
🚀

PR #600, opened 2022-07-26
Solution

yaroslav-lizogub:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
+  return grades.reduce((sum, prev) => sum + prev, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student [],
```
Oleksandr-Pustovit, line 25: _just a comment_
This space after type feels weird. previosly in code (line 20) you did not make it.
`(grades: number[])` 
It refers to all the next occurencies of `Student []`

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+        return order === 'asc'
+          ? Number(studentA.married) - Number(studentB.married)
+          : Number(studentB.married) - Number(studentA.married);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        return order === 'asc'
+          ? getAverageGrade(studentA.grades) - getAverageGrade(studentB.grades)
+          : getAverageGrade(studentB.grades) - getAverageGrade(studentA.grades);
```
Oleksandr-Pustovit, line 58: _comment_
`Default` is usually used to thow errors if switch case gets something it can not operate. May be It is better to make a separate case for averarageGrades.

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+
+        return order === 'asc'
+          ? studentA.name.localeCompare(studentB.name)
+          : studentB.name.localeCompare(studentA.name);
+
+      case SortType.Surname:
+
+        return order === 'asc'
+          ? studentA.surname.localeCompare(studentB.surname)
+          : studentB.surname.localeCompare(studentA.surname);
```
Oleksandr-Pustovit, line 43: These two have 100% identical logic. Switch allows combining several cases when put without brake command
It can be said about next 2 cases as well
```suggestion
      case SortType.Name:
      case SortType.Surname:

        return order === 'asc'
          ? studentA.surname.localeCompare(studentB.surname)
          : studentB.surname.localeCompare(studentA.surname);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+
+      case SortType.Age:
+        return order === 'asc'
+          ? studentA.age - studentB.age
+          : studentB.age - studentA.age;
```
Oleksandr-Pustovit, line 48: _comment_
I think it is better to use your Enum variable as keys in this case as you created it to be free of hardcoding.
```suggestion
          ? studentA[SortType.Age] - studentB[SortType.Age]
          : studentB[SortType.Age] - studentA[SortType.Age];
```

Oleksandr-Pustovit (changes requested):
Well done!
There are several comments to think and one thing to be fixed!

src/sortStudents.ts
yaroslav-lizogub: I'm not sure this is a good point

src/sortStudents.ts
yaroslav-lizogub: I'm not sure this is a good point

Oleksandr-Pustovit (approved):

src/sortStudents.ts
ivan-kadykalo: This is a good point, but to do this you need to use computed keys
For example
`studentA[sortBy].localCompare(studentB[sortBy])`

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
```
ivan-kadykalo, line 15: Do you have such key in your student object??
```suggestion
  AverageGrade = 'grades',
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
GJ🔥

PR #614, opened 2022-08-01
Solution

oleg-zolkiewskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    switch (sortBy) {
+      case SortType.Name:
+        if (order === 'asc') {
+          return (a.name).localeCompare(b.name);
+        }
+
+        return (b.name).localeCompare(a.name);
```
valerii-derkach, line 35: use ternary operator here instead of `if` and in all cases

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+
+      case SortType.Surname:
+        if (order === 'asc') {
+          return (a.surname).localeCompare(b.surname);
+        }
```
valerii-derkach, line 40:
```suggestion
      case SortType.Name:
      case SortType.Surname:
`  your code here
`
```

combine cases

valerii-derkach (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    case SortType.Name:
+
+      studentsList.sort((studentA: Student, studentB: Student) => {
+        return (order === 'asc')
+          ? studentA.name.localeCompare(studentB.name)
+          : studentB.name.localeCompare(studentA.name);
+      });
+      break;
+
+    case SortType.Surname:
```
alexandra-protyanova, line 37: combine this two cases, logic is the same

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+
+    case SortType.Age:
+      studentsList.sort((studentA: Student, studentB: Student) => {
+        return (order === 'asc')
+          ? studentA.age - studentB.age
+          : studentB.age - studentA.age;
+      });
+      break;
+
+    case SortType.Married:
```
alexandra-protyanova, line 53: combine these too

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      studentsList.sort((studentA: Student, studentB: Student) => {
+        return (order === 'asc')
+          ? +studentA.married - +studentB.married
+          : +studentB.married - +studentA.married;
+      });
+      break;
+
+    case SortType.AverageGrade:
+    default:
```
alexandra-protyanova, line 62: in default better to throw error that, data wasn't valid

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #621, opened 2022-08-17
TS_sort_students

Vladyslav317:


src/sortStudents.ts
```diff
@@ -1,16 +1,91 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+      copyStudents.sort((a: Student, b: Student) => {
+        if (order === 'asc') {
+          return a.name.localeCompare(b.name);
+        }
+
+        return b.name.localeCompare(a.name);
+      });
```
marina-tilniak, line 40: it is good case for using ternary operator

marina-tilniak (changes requested):

PR #625, opened 2022-08-19
Solution

aleksandra-prognimak:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+  order: SortOrder,
+): Student[] {
+  const copyStudents: Student[] = [...students];
+
+  copyStudents.sort((studentA: Student, studentB: Student) => {
+    switch (sortBy) {
+      case SortType.Name:
+        if (order === 'asc') {
+          return studentA.name.localeCompare(studentB.name);
+        }
```
tiserett, line 35:
```suggestion
        if (order === 'asc') {
          return studentA[sortBy].localeCompare(studentB[[sortBy]);
        }
```

sortBy contains string value, which can be used as a key inside square brackets, if you rewrite all code like this, you can connect case name && case surname, you can also connect case age && case married, I guess you should also rewrite all code using ternary operator

tiserett (changes requested):
you should make your code cleaner ;)

tiserett (approved):
gj

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+const aveGrade = (points: number[]): number => (
+  points.reduce((sum, point) => sum + point, 0) / points.length);
```
AlinaMatuschak, line 19:
```suggestion
const getAverageGrade = (points: number[]): number => (
  points.reduce((sum, point) => sum + point, 0) / points.length,
);
```

AlinaMatuschak (approved):

AlinaMatuschak (approved):

PR #626, opened 2022-08-19
Solution

oles-partyka:


src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
 
+    default:
+      studentsCopy.sort((a, b) => (
+        (a.grades
+          .reduce((sum, current) => sum + current, 0) / a.grades.length)
+        - (b.grades
+          .reduce((
+            sum, current,
+          ) => sum + current, 0) / b.grades.length)) * theSortOrder);
+      break;
```
Bohdan-Maksymiuk, line 59: why do we need this logic for default if we don't now what data will be passed in that case
lets throw an error if it will be unexpected type
```
throw new Error('Enter valid sort type');
```

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+      );
+      break;
+
+    case SortType.Age:
+      studentsCopy.sort((a, b) => (a.age - b.age) * theSortOrder);
+      break;
+
+    case SortType.Married:
+      studentsCopy.sort((a, b) => (+a.married - +b.married) * theSortOrder);
+      break;
```
Bohdan-Maksymiuk, line 49: there are a lot of repeating logic here 
we can group cases
and you can access property without explicitly typing property name
you can use `sort By`
`student1[sortBy].localeCompare(student2[sortBy])`

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #627, opened 2022-08-22
solution

Kulchevych:


tiserett (approved):
gj ;)

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
```
dellachaise, line 15: [Checklist point 2](https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md)
Mostly in real projects(at work) now commonly use this code rule.

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAverageGrade = (grades: number[]): number => {
+  return grades.reduce((prev, next) => prev + next, 0) / grades.length;
```
dellachaise, line 22: Better rename this way according to [method](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)
```suggestion
  return grades.reduce((previous, current) => previous + current, 0) / grades.length;
```

Or better more specific for this task
```suggestion
  return grades.reduce((sum, grade) =>sum + grade, 0) / grades.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const cloneStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? cloneStudents
+          .sort((prev, next) => prev[sortBy].localeCompare(next[sortBy]))
+        : cloneStudents
+          .sort((prev, next) => next[sortBy].localeCompare(prev[sortBy]));
```
dellachaise, line 39: Better use meaningful variable names. Rename for example to `firstStudent` and `secondStudent`.

The same below 👇

dellachaise (changes requested):
Great start 👍 

Just add changes according to comments 👇

src/sortStudents.ts
dellachaise: I checked, you could leave this way, it also would be ok. In our project we commonly use CAPS but in other - could be another option 👌 

But variable names - fix please 😉

dellachaise (approved):
Well done 👍

PR #630, opened 2022-08-23
Solution

VolodymyrChuyko:


oksana-logos-frontend (approved):
Good solution!

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
```
dellachaise, line 15: [Checklist point 2](https://github.com/mate-academy/ts_sort_students/blob/master/checklist.md)
  Mostly in real projects (at work) now commonly use this code rule.

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+): Student[] {
+  const studentsList: Student[] = [...students];
+  const getAverageGrade = (student: Student): number => {
+    const gradesSum = student.grades.reduce((sum, grade) => sum + grade, 0);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    return gradesSum / student.grades.length;
+  };
+  let sortFunction: (a: Student, b: Student) => number;
```
dellachaise, line 32: Better use meaningful variable names.
```suggestion
  let sortFunction: (firstStudent: Student, secondStudent: Student) => number;
```

dellachaise (changes requested):
Great job 👍 

Just add a few small changes 👇

src/sortStudents.ts
dellachaise: I checked, you could leave this way, it also would be ok. In our project we commonly use CAPS but in other - could be another option 👌 

But variable names - fix please 😉

VolodymyrChuyko: I just recently tried to change it and find out that these values are related to tested object keys. So it only works in lowercase. It was useful to learn this =)

dellachaise (approved):
Well done 👍  🚀

PR #632, opened 2022-08-24
solution

Githi54:


oksana-logos-frontend (approved):
Good job!

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAvarageGrade(grades: number[]): number {
+  return grades.reduce((a, b) => a + b, 0) / grades.length;
```
dellachaise, line 21: Better use meaningful variable names
 Rename this way according to [method](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)
```suggestion
  return grades.reduce((previous, current) => previous + current, 0) / grades.length;
```

Or better more specific for this task
```suggestion
  return grades.reduce((sum, grade) =>sum + grade, 0) / grades.length;
```

dellachaise (changes requested):
Great job 👍 

Just pay attention to variable names, it should be meaningful.

dellachaise (approved):
Well done :

PR #633, opened 2022-08-25
Develop

AlexanderKolomiiets:


AndreyJigijalkin (approved):
Great solution!)

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+const getAverage = (value : number[]) : number => value
+  .reduce((prev, curr) => prev + curr, 0) / value.length;
```
dellachaise, line 20: better use more meaningful variable names.
```suggestion
const getAverage = (grades : number[]) : number => grades
  .reduce((prev, curr) => prev + curr, 0) / value.length;
```
Instead of `prev, curr` better use `sum, grade` it would be more obvious for this function.

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+) : Student[] {
+  return [...students].sort((a, b) => {
```
dellachaise, line 27: The same about `a,b` - use more meaningfull variable names.

dellachaise (changes requested):
Great job 👍 

Just pay attention to variable names, it is important that they should be meaningful(all over the code).

PR #637, opened 2022-08-25
Task solution

sergii-nosachenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+
+    if (typeof paramA === 'string' && typeof paramB === 'string') {
+      diff = paramA.localeCompare(paramB);
+    }
+
+    if (typeof paramA === 'boolean' && typeof paramB === 'boolean') {
+      diff = Number(paramA) - Number(paramB);
+    }
+
+    if (Array.isArray(paramA) && Array.isArray(paramB)) {
```
AlexanderKolomiiets, line 44: Great decision to compare parameters by data type 💯

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsSorted = [...students];
+  const direction: -1 | 1 = order === 'asc' ? 1 : -1;
+
+  studentsSorted.sort((studA, studB): number => {
```
AlexanderKolomiiets, line 27:
```suggestion
  const direction: -1 | 1 = order === 'asc' ? 1 : -1;

  return [...students].sort((studA, studB): number => {
```

AlexanderKolomiiets: you can make it shorter

AlexanderKolomiiets (approved):
Brilliant !

src/sortStudents.ts
sergii-nosachenko: @AlexanderKolomiiets Thanks, added some optimizations)

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+        diff = Number(paramA) - Number(paramB);
+      }
+
+      if (Array.isArray(paramA) && Array.isArray(paramB)) {
+        const getAverage = (param: number[]): number => (
+          param.reduce((sum, num) => sum + num, 0) / param.length
+        );
 
+        diff = getAverage(paramA) - getAverage(paramB);
+      }
```
AlinaMatuschak, line 48: It should work just for grades array, not for all possible arrays

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+        diff = paramA - paramB;
+      }
+
+      if (typeof paramA === 'string' && typeof paramB === 'string') {
+        diff = paramA.localeCompare(paramB);
+      }
+
+      if (typeof paramA === 'boolean' && typeof paramB === 'boolean') {
+        diff = Number(paramA) - Number(paramB);
+      }
```
AlinaMatuschak, line 40: you can combine it. But this condition is hard to read, so maybe better use SortType here )

```suggestion
      if (typeof paramA === 'string' && typeof paramB === 'string') {
        diff = paramA.localeCompare(paramB);
      }

      if ((typeof paramA === 'boolean' && typeof paramB === 'boolean')
        || (typeof paramA === 'number' && typeof paramB === 'number')
      ) {
        diff = Number(paramA) - Number(paramB);
      }
```

AlinaMatuschak (changes requested):
If you have predetermined amount of options it is better to use switch/case structure instead of multiple if blocks

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+          (Number(paramA) - Number(paramB)) * direction
+        ),
+      );
+
+    case SortType.AverageGrade:
+      return [...students].sort(
+        ({ [sortBy]: paramA }, { [sortBy]: paramB }) => {
+          const getAverage = (param: number[]): number => (
+            param.reduce((sum, num) => sum + num, 0) / param.length
+          );
```
AlinaMatuschak, line 48: you can create this function outside the sortStudents. Your case for AverageGrade will be easier )

AlinaMatuschak (approved):
🔥

PR #645, opened 2022-08-30
Solution

Jamal-Amoian:


max-kravchenko (approved):
I guess this is a good code, but according to checklist I would pay attention to this field:
"Combine logic for sorting by fields with same data types."

like
case SortType.Name:
case SortType.Surname:
 return bla bla a[sortBy]; 

p.s. And I would rename a & b for smth like Student or Person

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      }
+
+      return tempStudents.sort((studentA, studentB) => {
+        return +studentB[sortBy] - +studentA[sortBy];
+      });
+
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        return tempStudents.sort(
+          (a, b) => calcSum(a[sortBy]) - calcSum(b[sortBy]),
```
dellachaise, line 60: Use meaningful variable names instead of `a, b`.

dellachaise (approved):
Well done 👍 

Just pay attention to variable names. It should be meaningful.

PR #659, opened 2022-08-30
solutions

KorniychukLesya:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+      studentsCopy
+        .sort((prev, next) => prev.name.localeCompare(next.name)
+        * orderSwitch);
+      break;
+    case SortType.Surname:
+      studentsCopy
+        .sort((prev, next) => prev.surname.localeCompare(next.surname)
+        * orderSwitch);
```
NataliyaAn, line 39: There is no need to write sort for each case. if they are the same. Like that
case SortType.Name:
case SortType.Surname:
 studentsCopy.sort(...);

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+      break;
+    case SortType.Age:
+      studentsCopy
+        .sort((prev, next) => (prev.age - next.age)
+        * orderSwitch);
+      break;
+    case SortType.Married:
+      studentsCopy
+        .sort((prev, next) => (+prev.married - +next.married)
+        * orderSwitch);
```
NataliyaAn, line 49: To convert to a number, it will be better using Number(). But if TS did not see an error, then everything is ok)

NataliyaAn (approved):
Good job!

src/sortStudents.ts
KorniychukLesya: Thank you. I completely agree with the comments. I fixed everything.

src/sortStudents.ts
KorniychukLesya: I fixed everything

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+      studentsCopy
+        .sort((prev, next) => (Number(prev.married) - Number(next.married))
+        * orderSwitch);
+      break;
+    default:
+      studentsCopy
+        .sort((prev, next) => (
+          ((prev.grades
+            .reduce((numPrev, numNext) => numPrev + numNext, 0))
+            / prev.grades.length)
```
dellachaise, line 52: you write this code twice, better create a helper function `getAverageGrades` for example for this purpose. And call it here.

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  const studentsCopy = [...students];
+  const orderSwitch: number = (order === 'asc') ? 1 : -1;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      studentsCopy
+        .sort((prev, next) => prev[sortBy].localeCompare(next[sortBy])
```
dellachaise, line 34: Variable names should be meaningful, for this task better use `firstStudent and secondStudent` for example, instead of `prev, next`.

dellachaise (changes requested):
Great job 👍 

Just fix 2 little issues 👇

KorniychukLesya:
Thank you! 🙂 Made all the changes.

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
+  return grades
+    .reduce((firstGreads, secondGreads) => firstGreads + secondGreads, 0)
+    / grades.length;
```
dellachaise, line 25: where you wrote `firstGrade` - there stored in reduce method `sum or total` and `secondGrade` - it is `currentGrade or simply grade`.
Rename)

dellachaise (approved):
Great job 🚀

PR #662, opened 2022-08-30
Solution

Mejeezz:
Had a little problem, but it didn't impact on test passing .  What is it? Should i ignore it?
photo_2022-08-30_23-50-40


Ohiiko (approved):
Good job!

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? listOfStudents
+          .sort((studentOne, studentTwo) => (
+            +studentOne[sortBy] - +studentTwo[sortBy]))
+        : listOfStudents
+          .sort((studentOne, studentTwo) => (
+            +studentTwo[sortBy] - +studentOne[sortBy]));
```
dellachaise, line 51: In what case do you have a problem from your screen? If in this case - I think you already resolved it by using `+` and this way boolean from `married` become a number. I think the problem was because of it.

dellachaise (approved):
Great job 🚀

src/sortStudents.ts
dellachaise: Already saw your second commit. It really was because of `married` type, you also could use `Number()` instead of `+` 😉  but `+` also ok)

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
```
dellachaise, line 53: It shouldn't be a default case. Default - could be some error. But here - the separate case. Rewrite please.

Mejeezz: Done! 👍

PR #664, opened 2022-08-31
Solution

OlgaOchichenko:


anyakrainiuk (approved):
Good Job!

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? arrSortStudents.sort((a, b) => (
+          getAverageGrades(a.grades) - getAverageGrades(b.grades)))
+        : arrSortStudents.sort((a, b) => (
+          getAverageGrades(b.grades) - getAverageGrades(a.grades)));
+
+    default:
+      break;
```
AlinaMatuschak, line 60: use default case for throwing errors or default value )

AlinaMatuschak (changes requested):
Combine logic for sorting by fields with same data types. It will be easier if on SortType enum your values have lowercase case. Then you could use a[sortBy]

AlinaMatuschak (changes requested):
still point 3 from checklist

AlinaMatuschak (approved):
🔥🔥

PR #665, opened 2022-08-31
Implement Student Sorting func in TS

max-kravchenko:
Implement Student Sorting func in TS


Jamal-Amoian (approved):
Perfectly!

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
+function getAverage([...args]: number[]): number {
+  return args.reduce((x, y) => x + y, 0) / args.length;
```
dellachaise, line 2: better use more meaningfull variable names. 
For this task `sum and grade` instead of `x and y` , if you want to reuse this function - `total or sum and current`.

dellachaise (approved):
Well done 👍 

Just pay attention to helper function variable names.

src/sortStudents.ts
max-kravchenko: Thanks!!

PR #675, opened 2022-08-31
Solution

Roman-Artiukh:


ostUp (approved):
🚀

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAverage = (value : number[]): number => value
+  .reduce((a, b) => a + b, 0) / value.length;
```
dellachaise, line 25: Better use more meaningful variables, for example.
```suggestion
const getAverage = (grades : number[]): number => grades
  .reduce((sum, grade) => sum + grade, 0) / value.length;
```
or this way if you want to reuse this function in other project 👇 
```suggestion
const getAverage = (dataArray : number[]): number => dataArray
  .reduce((sum, current) => sum + current, 0) / value.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] | number {
   // write your function
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? [...students].sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
```
dellachaise, line 38: Better create a variable instead of `[...students]` and use it all over the code.

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+        : [...students].sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```
dellachaise, line 39: Use meaningful variable names instead of `a, b` - all over the code.

dellachaise (changes requested):
Great job 👍 

Just pay attention to variable names it should be meaningful.

dellachaise (approved):
Well done 👍

PR #678, opened 2022-08-31
TS solution Sort Students

wfurs0w:


OleksiiPustovit (approved):
Good job 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverage(grades: number[]): number {
+  return grades.reduce((sum, value) => sum + value) / grades.length;
```
dellachaise, line 22: if you already use here `grades` to more meaningful would be rename `value` too.
```suggestion
  return grades.reduce((sum, grade) => sum + grade) / grades.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  order: SortOrder,
+): Student[] {
+  const copyOfStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? copyOfStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : copyOfStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```
dellachaise, line 37: Variable names should be meaningful all over the code. Better use `firstStudent and secondStudent` for example instead of `a and  b`.

dellachaise (changes requested):
Great job 👍 

Just pay attention to variable names - they should be meaningful.

src/sortStudents.ts
wfurs0w: Initially, I called them that)) but the writing seemed to me too long

src/sortStudents.ts
dellachaise: Still actual)

dellachaise (approved):
Well done 👍

PR #689, opened 2022-09-01
Sort students

Sasha-Gordijchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+        return -1;
+      }
+
+      if (first[sortBy] < second[sortBy]) {
+        return 1;
+      }
 
+      return 0;
+    });
+  }
```
plemiannik, line 60: image
anyway u must use switch instead of if`s, cos checklist

plemiannik (changes requested):
excellent

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+        if (first[sortBy] < second[sortBy]) {
+          return 1;
+        }
+
+        return 0;
+      });
+      break;
 
+    default:
+      return sortedStudens;
```
plemiannik, line 64: interesting solution, but why u need break? also i think u can do just any case with 1 ternar operator where u return that case if order is same and return same but .reverse() if not. anyway, like ur solution

plemiannik (approved):
✈️

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  switch (order) {
+    case 'asc':
+      sortedStudens.sort((first: Student, second: Student) => {
+        if (first[sortBy] < second[sortBy]) {
+          return -1;
+        }
+
+        if (first[sortBy] > second[sortBy]) {
+          return 1;
+        }
```
dellachaise, line 43: in such case better use the ternary operator. And if you already have return you shouldn't write `break`.

dellachaise (changes requested):
Interesting solution and good job 👍 

But according to the task, we should sort by types of SortType, not order. And then in each case - sort would be a different relay to order.
Rewrite the code, please.

And pay attention to variable names (comments).

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+    switch (order) {
+      case 'asc':
+        return firstStudent[sortBy] < secondStudent[sortBy]
+          ? -1
+          : 1;
+
+      case 'desc':
+        return firstStudent[sortBy] > secondStudent[sortBy]
+          ? -1
+          : 1;
```
dellachaise, line 44: This point in task mean that you should build your code using kind of this logic:

Screenshot 2022-09-02 at 10 30 32

```suggestion
    switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return order === 'asc'
        ? ...
        : ...

    case SortType.Age:
    case SortType.Married:
      return order === 'asc'
        ? ...
        : ...

```

dellachaise (changes requested):
The switch logic is still wrong. Rewrite, please, acoording to more detailed comment 👇

AndriiRut (approved):
🚀

PR #691, opened 2022-09-01
sortStudent created

Reptiloid044:
Hi!


src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+      return order === 'asc'
+        ? copyStudents.sort((first, second) => first[sortBy] - second[sortBy])
+        : copyStudents.sort((first, second) => second[sortBy] - first[sortBy]);
+
+    case SortType.Married:
+      return order === 'asc'
+        ? copyStudents.sort((first, second) => (
+          Number(first[sortBy]) - Number(second[sortBy])))
+        : copyStudents.sort((first, second) => (
+          Number(second[sortBy]) - Number(first[sortBy])));
```
OleksiiPustovit, line 49: You can also combine these blocks
```suggestion
  case SortType.Age:
    case SortType.Married:
      return order === 'asc'
        ? copyStudents.sort((first, second) => first[sortBy] - second[sortBy])
        : copyStudents.sort((first, second) => second[sortBy] - first[sortBy]);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+          calculateAvarageGrade(first[sortBy]) - calculateAvarageGrade(second[sortBy])))
+        : copyStudents.sort((first, second) => (
+          // eslint-disable-next-line
+          calculateAvarageGrade(second[sortBy]) - calculateAvarageGrade(first[sortBy])));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return copyStudents;
+  }
```
OleksiiPustovit, line 62: And here I propose to display an error
```suggestion
   default:
      throw new Error('Unknown sort type');
```

OleksiiPustovit (changes requested):
Good job, but I have a couple of suggestions))

src/sortStudents.ts
Reptiloid044: Yeah, I understood what are you talking about. But in my opinion, it is looking not as obvious as my solution. 
I think that compiling boolean to number will be not correct in this case... So I decided to do by this way =)

OleksiiPustovit (approved):
🚀

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+      return order === 'asc'
+        ? copyStudents.sort((first, second) => (
+          Number(first[sortBy]) - Number(second[sortBy])))
+        : copyStudents.sort((first, second) => (
+          Number(second[sortBy]) - Number(first[sortBy])));
 
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? copyStudents.sort((first, second) => (
+          // eslint-disable-next-line
```
AlinaMatuschak, line 54: 😂😂😂
```suggestion
```

AlinaMatuschak (approved):
Don't forget about code style )

PR #692, opened 2022-09-01
Sort students

Desertiks:


src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order:SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  function getAverageGrade(grades: number[]): number {
+    const gradesSum = grades.reduce((s: number, c: number) => s + c);
```
Reptiloid044, line 28: Better to use a more understandable attributes name
See more here: [https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce](url)

Reptiloid044 (approved):
Nice job 🦾

src/sortStudents.ts
Desertiks: I wrote these names because of the line limit(

Reptiloid044: Yeah, I know that it sometimes is hard to fix when the linter is not satisfied, but you know, you should write understandable code and you should adhere to the code architecture that was made by the author =)
And, there are a lot of solutions to make code as requested by the author 🦾

src/sortStudents.ts
```diff
@@ -25,7 +25,9 @@ export function sortStudents(
   const studentsCopy: Student[] = [...students];
 
   function getAverageGrade(grades: number[]): number {
-    const gradesSum = grades.reduce((s: number, c: number) => s + c);
+    const gradesSum = grades.reduce((storage: number, current: number) => {
```
Reptiloid044, line 28: Nice 🤘

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+        return (order === 'asc'
+          ? firstStudent[sortBy].localeCompare(secondStudent[sortBy])
+          : firstStudent[sortBy].localeCompare(secondStudent[sortBy]));
+
+      default:
+        return (order === 'asc'
+          ? getAverageGrade(firstStudent.grades)
+        - getAverageGrade(secondStudent.grades)
+          : getAverageGrade(secondStudent.grades)
+        - getAverageGrade(firstStudent.grades));
```
AlinaMatuschak, line 54: It should work just for AverageGrade. You can use the default case for throwing errors or returning a default value )

AlinaMatuschak (approved):

PR #699, opened 2022-09-02
ts_sort_students

yurii-kozlov:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+    case SortType.AverageGrade:
+      return studentsCopy.sort((student1, student2) => (order === 'asc'
+        ? getAverageGrade(student1[sortBy]) - getAverageGrade(student2[sortBy])
+        : getAverageGrade(student2[sortBy]) - getAverageGrade(student1[sortBy])
+      ));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return [];
```
alexandra-protyanova, line 53: it is better to throw error in default case

alexandra-protyanova (approved):

PR #704, opened 2022-09-06
create ts sorter

DominoSl:


src/sortStudents.ts
```diff
@@ -1,16 +1,80 @@
...
+      ));
+    }
+
+    case SortType.Surname: {
+      return studentsCopy.sort((
+        studentA, studentB,
+      ) => (order === 'asc'
+        ? studentA.surname.localeCompare(studentB.surname)
+        : studentB.surname.localeCompare(studentA.surname)
+      ));
```
Bohdan-Maksymiuk, line 46: we can group similar cases and use smth like
```
student1[sortBy].localeCompare(student2[sortBy])
```

Bohdan-Maksymiuk (changes requested):

AlisaVlasova (approved):

PR #706, opened 2022-09-13
Develop

HusakOleh:


David-Shahmatov (approved):
🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,128 @@
...
+      });
+      break;
+    case SortType.AverageGrade:
+      resArr.sort((a: Student, b: Student): number => {
+        const averageStudentGrade = (student: Student):number => {
+          return student
+            .grades.reduce((accum: number, grade: number): number => {
+              return accum + grade;
+            }, 0) / student.grades.length;
+        };
```
dellachaise, line 100: Write this function outside `swith case` you could write it after types.

src/sortStudents.ts
```diff
@@ -1,16 +1,128 @@
...
+      });
+      break;
+    case SortType.Surname:
+      resArr.sort((a: Student, b: Student): number => {
+        if (sortOrder) {
+          return a.surname.localeCompare(b.surname);
+        }
+
+        return b.surname.localeCompare(a.surname);
+      });
```
dellachaise, line 45: You could write it with one case because the code is nearly the same.

And use ternary operator instead of `if` all over the code.
```suggestion
    case SortType.Name:
    case SortType.Surname:
    .....
```

src/sortStudents.ts
```diff
@@ -1,16 +1,128 @@
...
+): Student[] {
+  const resArr: Student[] = [...students];
+  const sortOrder: boolean = order === 'asc';
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      resArr.sort((a: Student, b: Student): number => {
+        if (sortOrder) {
```
dellachaise, line 31:
```suggestion
        return order === 'asc'
        ? first result
        : second result;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,128 @@
...
+
+        if (Number(a.age) < Number(b.age)) {
+          if (!sortOrder) {
+            res = 1;
+          } else {
+            res = -1;
+          }
+        }
+
+        return res;
```
dellachaise, line 67: you already have numbers here, don't need to use `Number`.

You could simply use the ternary operator, without `res`. Simplify this way all over the code.
```suggestion
   return order === 'asc' 
   ? a.age - b.age
   : b.age - a.age;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,128 @@
...
+  order: SortOrder,
+): Student[] {
+  const resArr: Student[] = [...students];
+  const sortOrder: boolean = order === 'asc';
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      resArr.sort((a: Student, b: Student): number => {
```
dellachaise, line 30: `a and b` - not a good variable names. Variable names should be meaningful. Better use `firstStudent and secondStudent` for example.

dellachaise (changes requested):
Good start 👍 

Just pay attention to comments and make your code cleaner.

src/sortStudents.ts
```diff
@@ -1,16 +1,98 @@
...
+
+        if (firstStudent.age > secondStudent.age) {
+          res = sortOrder ? 1 : -1;
+        }
+
+        if (firstStudent.age < secondStudent.age) {
+          res = !sortOrder ? 1 : -1;
+        }
+
+        return res;
```
dellachaise, line 56: You don't need `res` variable, you could simply write this way 👇 
```suggestion
        return sortOrder
          ? firstStudent.age - secondStudent.age
          : secondStudent.age - firstStudent.age;
```

for `SortType.Married and SortType.AverageGrade` the same but addapt it ☝️

dellachaise (approved):
Well done 👍 

Just make it simpler one more time 👇

PR #707, opened 2022-09-15
Solution

Viktorianeimesh:


src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+
+      case SortType.Surname:
+        return newStudents.sort((a, b) => (a.surname.localeCompare(b.surname)));
+
+      case SortType.Age:
+        return newStudents.sort((a, b) => (a.age - b.age));
+
+      case SortType.Married:
+        return newStudents.sort((a, b) => (+a.married - +b.married));
+
```
SanyaBratashchuk, line 45: You can group Name with Surname and Married with Age

SanyaBratashchuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+        return newStudents.sort((student1, student2) => (
+          +student1[sortBy] - (+student2[sortBy])));
+
+      case SortType.AverageGrade:
+        return newStudents
+          .sort((student1, student2) => (
+            getAvgGrade(student1.grades) - getAvgGrade(student2.grades)));
+
+      default:
+        return [];
```
SanyaBratashchuk, line 50:
```suggestion
        return newStudents;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+        return newStudents.sort((student1, student2) => (
+          +student2[sortBy] - (+student1[sortBy])));
+
+      case SortType.AverageGrade:
+        return newStudents
+          .sort((student1, student2) => (
+            getAvgGrade(student2.grades) - getAvgGrade(student1.grades)));
+
+      default:
+        return [];
```
SanyaBratashchuk, line 72:
```suggestion
        return newStudents;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+    }
+  }
+
+  return newStudents;
```
SanyaBratashchuk, line 76:
```suggestion
```

SanyaBratashchuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+        return newStudents
+          .sort((student1, student2) => (
+            getAvgGrade(student2.grades) - getAvgGrade(student1.grades)));
+
+      default:
+        return newStudents;
+    }
+  }
+
+  return newStudents;
```
TarasHoliuk, line 76: taking into account that you already covered all possible cases (including `default: return newStudents;`) you could remove it - this line of code will never run

TarasHoliuk (changes requested):
good job, small fix is needed

src/sortStudents.ts
Viktorianeimesh: If I remove the line  return newStudents at the end of the code I receive 2 errors:
- Expected to return a value at the end of function 'sortStudents'.
- Function lacks ending return statement and return type does not include 'undefined'.

TarasHoliuk: Ok, got it
Probably, that's because in this case, linter doesn't know that `asc` and `desc` are all possible cases
Then leave it as it is, approved

TarasHoliuk (approved):

PR #709, opened 2022-09-19
Done

Oleksii-Blyzniuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+
+      return sortedStudents.sort((a, b) => {
+        return b[sortBy].localeCompare(a[sortBy]);
+      });
+
+    case SortType.Age:
+    case SortType.Married:
+      if (order === 'asc') {
+        return sortedStudents.sort((a, b) => {
+          return +a[sortBy] - +b[sortBy];
```
liliia-livandovskaya, line 48:
```suggestion
          return Number(a[sortBy]) - Number(b[sortBy]);
```
Number() is more obvious way

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      if (order === 'asc') {
+        return sortedStudents.sort((a, b) => {
```
liliia-livandovskaya, line 35: don't use a, b. Rename variables to understand what they contain of

liliia-livandovskaya (approved):
🔥

PR #712, opened 2022-09-20
first solution

fatihsamur:
need some refactoring for reducing if-else usage


Quelaan1 (approved):
Too complicated work I guess

ihor-jpeg (approved):

PR #720, opened 2022-09-26
Solution

AMahalias:


src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): object[] {
```
qoa11a, line 20: According to the task function must return a new array with students. So you can specify it here instead of object[].

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+  const sortedStudents = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+      return (order === 'asc')
```
qoa11a, line 25: You can check for order inside sort method.
`sortedStudents.sort((studentA, studentB) => {
  return order === 'asc' ? ascLogic : descLogic;
})`

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+        ? sortedStudents.sort(
+          (studentA, studentB) => studentA.name.localeCompare(studentB.name),
+        )
+        : sortedStudents.sort(
+          (studentA, studentB) => studentB.name.localeCompare(studentA.name),
+        );
+
+    case SortType.Surname:
+      return (order === 'asc')
```
qoa11a, line 34: Same here and elsewhere.

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+        : sortedStudents.sort((studentA, studentB) => +studentB.married
+        - (+studentA.married));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    case SortType.AverageGrade:
+      return (order === 'asc')
+        ? sortedStudents.sort(
+          (studentA, studentB) => (studentA.grades
+            .reduce((sum, x) => sum + x)) / studentA.grades.length
```
qoa11a, line 62: This code is unreadable. I recommend you to create helper function to find average value from an array of numbers.

qoa11a (changes requested):

SanyaBratashchuk (approved):

PR #722, opened 2022-09-27
Develop

SalehAkaJim:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+        : copyStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+
+    case SortType.Married:
+    case SortType.Age:
+      return order === 'desc'
+        ? copyStudents.sort((a, b) => +b[sortBy] - +a[sortBy])
+        : copyStudents.sort((a, b) => +a[sortBy] - +b[sortBy]);
 
+    case SortType.AverageGrade:
+      return order === 'desc'
```
mgrinko, line 47: You could do 1 reverse after switch instead of checking this condition 3 times

mgrinko (approved):

src/sortStudents.ts
SalehBUD: Hopefully, it's now fixed 😅

PR #723, opened 2022-09-27
sort complete

Yevhen-Pavlovskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades:number[]):number {
```
qoa11a, line 24:
```suggestion
function getAverageGrade(grades: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  return grades.reduce((sum, currentValue) => sum
+  + currentValue, 0) / grades.length;
+}
 
-export function sortStudents(students, sortBy, order) {
+export function sortStudents(students: Student, sortBy: SortType, order: SortOrder): Student[] {
   // write your function
+  const copy: Student[] = Object.assign([], students);
```
qoa11a, line 31: It's correct way to copy an array but you can also do it like this
`const copy = [...students];`

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
```
qoa11a, line 36: You can move check for order inside method sort so there will be less repetitive code.

qoa11a (changes requested):

liliia-livandovskaya (changes requested):
Remove all the comments before sending the task for review

Bohdan-Maksymiuk (changes requested):
you duplicate a lot of code
make 1 sort and switch by cases into sort callback

PR #728, opened 2022-09-29
Solution

AntonVaida:


src/sortStudents.ts
```diff
@@ -1,16 +1,114 @@
...
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students:Student[],
+  sortBy:SortType,
+  order:SortOrder,
+):Student[] {
+  const studentList:Student[] = students.map(
+    (student:Student) => ({ ...student }),
+  );
```
Anna-Tkachenko, line 28: you can do it simpler
```suggestion
  const studentList:Student[] = [...students];
```

Anna-Tkachenko (changes requested):
fix comments and check point 3 in checklist

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
 
 export interface Student {
-  // describe Student interface
+  name:string,
+  surname: string,
+  age: number,
+  married:boolean,
+  grades:number[],
```
OleksiiBlezniuk, line 7:
```suggestion
  name: string,
  surname: string,
  age: number,
  married: boolean,
  grades: number[],
```

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
+}
+
+function avaregeGrades(grades:number[]):number {
+  return grades.reduce((sum, value) => sum + value) / grades.length;
 }
```
OleksiiBlezniuk, line 20: Ok now, but it's more logical to create helper functions that can be reused for other values. So it's more logical to create more generic names for it instead of specific. For example this function could be named as `getAverage` and accept array of generic `numbers` instead of specific `grades` to be reusable for other arrays too, not only for grades 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students:Student[],
+  sortBy:SortType,
+  order:SortOrder,
```
OleksiiBlezniuk, line 28:
```suggestion
  students: Student[],
  sortBy: SortType,
  order: SortOrder,
```

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+):Student[] {
+  const studentList:Student[] = [...students];
```
OleksiiBlezniuk, line 30:
```suggestion
): Student[] {
  const studentList: Student[] = [...students];
```

OleksiiBlezniuk: Where have you lost spaces? 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+            }
+
+            if (a.married) {
+              return 1;
+            }
+
+            return -1;
+          },
+        );
+      }
```
OleksiiBlezniuk, line 71: Ok, but isn't be the same if we'll convert boolean to numbers? If we'll do it that way we'll be able to combine this case with previous and make our code much shorter

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+          if (a.married) {
+            return -1;
+          }
 
+          return 1;
+        },
+      );
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        return studentList.sort(
```
OleksiiBlezniuk, line 88: Also we can move switch/case inside `sort` method and avoid repeating of this line of code every time (we'll just return needed compare result

OleksiiBlezniuk (changes requested):

Denchepornyuk (approved):

PR #730, opened 2022-09-30
Develop

grafit-off:


vladimir-dublya (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function getAvarageGrades(grades: number[]): number {
```
silchyn, line 22: avarage -> average

silchyn (changes requested):
Here it is convenient and better to use switch-case

src/sortStudents.ts
grafit-off: I have only two if blocks, switch case is looks redundant in my case, isn't?

grafit-off: My bad

PR #732, opened 2022-10-02
implemented

ArturNagaycev:


src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentA.age - studentB.age;
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentB.age - studentA.age;
+        });
+      }
+      break;
```
Vladosik1337, line 69: I think you don`t need `else` if you have return

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return +studentA.married - +studentB.married;
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return +studentB.married - +studentA.married;
+        });
+      }
+      break;
```
Vladosik1337, line 81: And here

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return getAvgGrade(studentA.grades) - getAvgGrade(studentB.grades);
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return getAvgGrade(studentB.grades) - getAvgGrade(studentA.grades);
+        });
+      }
+      break;
```
Vladosik1337, line 93: And here

Vladosik1337 (approved):
Nice work)

src/sortStudents.ts
ArturNagaycev: this return is in internal function, not in the else block

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+    case SortType.Surname:
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentA.surname.localeCompare(studentB.surname);
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentB.surname.localeCompare(studentA.surname);
+        });
+      }
```
OleksiiBlezniuk, line 56: Same logic, just combine cases
```
case A:
case B:
  return 123;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+
+    case SortType.Age:
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentA.age - studentB.age;
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return studentB.age - studentA.age;
+        });
```
OleksiiBlezniuk, line 67: Same here, no need to duplicate `sort`

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+    case SortType.Married:
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return +studentA.married - +studentB.married;
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return +studentB.married - +studentA.married;
+        });
+      }
```
OleksiiBlezniuk, line 80: And here

src/sortStudents.ts
```diff
@@ -1,16 +1,100 @@
...
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return getAvgGrade(studentA.grades) - getAvgGrade(studentB.grades);
+        });
+      } else {
+        copyStudents.sort((studentA: Student, studentB: Student): number => {
+          return getAvgGrade(studentB.grades) - getAvgGrade(studentA.grades);
+        });
+      }
```
OleksiiBlezniuk, line 92: And here

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+      break;
+
+    case SortType.Married:
+      copyStudents.sort((studentA: Student, studentB: Student): number => {
+        return order === 'asc'
+          ? +studentA[sortBy] - +studentB[sortBy]
+          : +studentB[sortBy] - +studentA[sortBy];
+      });
+      break;
```
OleksiiBlezniuk, line 58: Now this sort conditions looks similar 🙂 You can use one sort for both cases
Also `Number()` is more readable then `+`

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
ArturNagaycev: really. thanks

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+    case SortType.AverageGrade:
+      copyStudents.sort((studentA: Student, studentB: Student): number => {
+        return order === 'asc'
+          ? getAvgGrade(studentA[sortBy]) - getAvgGrade(studentB[sortBy])
+          : getAvgGrade(studentB[sortBy]) - getAvgGrade(studentA[sortBy]);
+      });
+      break;
+
+    default:
+      throw new Error('Unknown sort type');
```
OleksiiBlezniuk, line 62: Or you can return unsorted or empty array here
As for me it's more logical action, probably we don't want to have errors because of wrong type in sorting function 🙂

OleksiiBlezniuk (approved):
Good

PR #736, opened 2022-10-03
Solution

vladimir-dublya:


src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+          : secondPerson.name.localeCompare(firstPerson.name)
+      ));
+      break;
+
+    case SortType.Surname:
+      studentsCopy.sort((firstPerson: Student, secondPerson: Student) => (
+        order === 'asc'
+          ? firstPerson.surname.localeCompare(secondPerson.surname)
+          : secondPerson.surname.localeCompare(firstPerson.surname)
+      ));
```
grafit-off, line 43: These two looks similar, you can try to unite them

grafit-off (approved):
GJ

silchyn (changes requested):
Good point above

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
+
+  const nameOrSurname = sortBy === SortType.Name ? 'name' : 'surname';
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      studentsCopy.sort((firstPerson: Student, secondPerson: Student) => (
```
naziamoff, line 34: you repeat this logic in each case of the switch. You should avoid repeated logic by all costs.
You can implement the `switch` inside the sorting function and it will be easier to avoid duplicated fragments of the code

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
+
+  const nameOrSurname = sortBy === SortType.Name ? 'name' : 'surname';
```
naziamoff, line 29: you don't need this variable. You should take a look at enums theory again and you will figure out how to remove it

naziamoff (changes requested):

naziamoff (changes requested):
> you repeat this logic in each case of the switch. You should avoid repeated logic by all costs.
You can implement the switch inside the sorting function and it will be easier to avoid duplicated fragments of the code

And [pay attention to the naming of your commits](https://mateacademy.slack.com/archives/C03REBVA3AN/p1665649990424539)

Denchepornyuk (approved):

PR #737, opened 2022-10-03
students have been sorted

illia-sosnytskyi:


Oleksandra-Shevchenko (approved):
LGTM 💯

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function calculateAverageGrade(student: Student): number {
+  return student.grades
+    .reduce((acc: number, prev: number) => acc + prev, 0)
+    / student.grades.length;
+}
```
OleksiiBlezniuk, line 24: Ok now, but it's more logical to create helper functions that can be reused for other values. So it's more logical to create more generic names for it instead of specific. For example this function could be named as getAverage and accept only array of numbers to be reusable for other arrays too, not only for student 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const outputArray: Student[] = [...students];
+
+  switch (sortBy) {
+    case 'age':
```
OleksiiBlezniuk, line 34: Avoid using magic string, you already have `enum` for that purpose

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+    case 'married':
+      outputArray.sort((currentStudent, previousStudent) => (
```
OleksiiBlezniuk, line 36: Also you can move switch/case inside `sort` method and avoid writing this method in every case

OleksiiBlezniuk (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #738, opened 2022-10-03
sort students task

TheVovchik:


src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+    acc + curr
+  ), 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsList: Student[] = JSON.parse(JSON.stringify(students));
```
grafit-off, line 30: Wow 🔥

grafit-off (approved):
GJ 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+          if (sortBy === SortType.AverageGrade) {
+            return order === 'asc'
+              ? avGrades(studA[sortBy]) - avGrades(studB[sortBy])
+              : avGrades(studB[sortBy]) - avGrades(studA[sortBy]);
+          }
 
+          return order === 'asc'
+            ? Number(studA[sortBy]) - Number(studB[sortBy])
+            : Number(studB[sortBy]) - Number(studA[sortBy]);
+        });
```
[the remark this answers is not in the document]
TheVovchik: because we sort numbers and in case of grades we need to calculate average, so we need to know when it's Grades what we sort. If you want so strong that I create separate case ok.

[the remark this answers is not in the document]
OleksiiBlezniuk: Ok, logic is clear, but still there is no need to create nested `if` check for type even if values which we used for sorting has same data types. We already have switch/case block where we check what SortType is passed, this `if` check inside case is just an unnecessary additional action to perform 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
 
+    case SortType.AverageGrade:
+      return studentsList
+        .sort((studA: Student, studB: Student) => {
+          return order === 'asc'
+            ? avGrades(studA[sortBy]) - avGrades(studB[sortBy])
+            : avGrades(studB[sortBy]) - avGrades(studA[sortBy]);
+        });
```
OleksiiBlezniuk, line 57: Btw, we could move this switch/case inside sort method, and there would be no need to write it every time 🙂

OleksiiBlezniuk (approved):

src/sortStudents.ts
TheVovchik: really :) God's ways are unfathomable )))))

PR #741, opened 2022-10-03
Sort Students TS

Anna-Borodaienko:


src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+    .reduce((acc: number, current: number) => acc + current) / (grades.length);
 }
 
-// create SortOrder type
-export type SortOrder;
+export
+function sortStudents(students: Student[], sortBy: SortType, order: SortOrder)
+  : Student[] {
+  const sortedStudents: Student[] = students
+    .map((person: Student) => ({ ...person }));
```
DenysKleimenov, line 52: In my opinion, destructuring would be clearer.

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  first: string,
+  second: string,
+  order: string,
+): number {
+  if (order === 'asc') {
+    return first.localeCompare(second);
+  }
+
+  return second.localeCompare(first);
+}
```
DenysKleimenov, line 29: I don't think it's necessary to create a separate function for strings, we use this function only once, so it would be good if you just used ternary operators, because we have only two options.

DenysKleimenov (approved):
LGTM

silchyn (changes requested):
Good job, but here it is convenient and better to use the switch-case

silchyn (changes requested):
`return sortedStudents.sort((first: Student, second: Student) => {` is repeated several times, so it would be nice if you restructure your code so that switch-case is inside the compare func

silchyn (approved):
Cool!

PR #742, opened 2022-10-03
Solution

Apollonia5:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+    case SortType.Name:
+    case SortType.Surname:
+      studentsCopy.sort((currStudent, nextStudent) => (
+        order === 'asc'
+          ? currStudent[sortBy].localeCompare(nextStudent[sortBy])
+          : nextStudent[sortBy].localeCompare(currStudent[sortBy])
+      ));
+      break;
+
+    case SortType.Age:
```
Yevheniya-Deryka, line 36: Maybe, it'll be better to combine this case with the next one (they are not actually the same, so I'm not sure about this suggestion)

Yevheniya-Deryka (approved):
GJ!

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+      ));
+      break;
+
+    case SortType.AverageGrade:
+      studentsCopy.sort((currStudent, nextStudent) => {
+        const currStudentAver = currStudent[sortBy]
+          .reduce((acc, curr) => acc + curr, 0) / currStudent[sortBy].length;
+        const nextStudentAver = nextStudent[sortBy]
+          .reduce((acc, curr) => acc + curr, 0) / nextStudent[sortBy].length;
+
```
Anna-Tkachenko, line 51: let's create separate function for calculating average grade value

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #744, opened 2022-10-03
Solution v01

anton-iskryk:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy]
+      ));
+
+    case SortType.Married:
+      return sortedStudents.sort((a: Student, b: Student) => (
+        order === 'asc'
+          ? +a[sortBy] - +b[sortBy]
+          : +b[sortBy] - +a[sortBy]
+      ));
```
artem-nasadchuk, line 52: You can use one sort for both cases. Also, Number() is more readable than +.

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  return grades
+    .reduce((acc: number, curr: number): number => acc + curr) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents: Student[] = [...students];
```
artem-nasadchuk, line 30: I think no need to specify the type here

artem-nasadchuk (approved):

src/sortStudents.ts
OleksiiBlezniuk, line 52: You used same sorting for both this cases, let's combine them
And use `Number` instead of `+`, it's more readable

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+    case SortType.Married:
+      return sortedStudents.sort((a: Student, b: Student) => (
+        order === 'asc'
+          ? +a[sortBy] - +b[sortBy]
+          : +b[sortBy] - +a[sortBy]
+      ));
+
+    case SortType.AverageGrade:
+      return sortedStudents.sort((a: Student, b: Student) => (
```
OleksiiBlezniuk, line 55: Btw, you can just move switch/case to sort method, no need to write it every time 🙂

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
OleksiiBlezniuk: You can skip type assigning here, it's understandable what value stored in this constant

OleksiiBlezniuk (approved):
GJ!

PR #745, opened 2022-10-03
Solution

iryna-borniak:


src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
+      });
+      break;
+
+    case SortType.Surname:
+      studentsCopy.sort((prev: Student, next: Student) => {
+        return order === 'asc'
+          ? prev[sortBy].localeCompare(next[sortBy])
+          : next[sortBy].localeCompare(prev[sortBy]);
+      });
+      break;
```
korniienkokarina, line 45: i think you can combine these conditions like these:
case 1:
case 2:
sort logic

src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
 
+    default:
+      return studentsCopy;
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return studentsCopy;
```
korniienkokarina, line 75: you can omit this as you return the same in default case

korniienkokarina (commented):
looks good and logical 👍🏻, 
left few comments to shorten the code 😊

src/sortStudents.ts
Rina07: thx) that's what I tried to do, just had an issue with linter demanding a 'break' before the 'case, but looks like with 'return' in 'cases', I don't have this issue anymore)

src/sortStudents.ts
Rina07: yep, with returns in every case I don't need one )

korniienkokarina (approved):

naziamoff (changes requested):
Well done, Iryna! 🔥 
Let's make it event better.

Now you repeat `studentsCopy.sort((prev: Student, next: Student)` a bunch of times - we should avoid the duplicated code fragments. Try to implement the `switch` inside the sorting callback

iryna-borniak:
> Well done, Iryna! fire Let's make it event better.
> 
> Now you repeat `studentsCopy.sort((prev: Student, next: Student)` a bunch of times - we should avoid the duplicated code fragments. Try to implement the `switch` inside the sorting callback

Thx!) pls, check this. Had to change the default case, as the callback expects a number

naziamoff (approved):
Well done 🚀

PR #746, opened 2022-10-03
SortStudents solution using TS

Yevheniya-Deryka:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  return grades.reduce(
+    (sum: number, grade: number) => sum + grade,
+  ) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): object[] {
```
Apollonia5, line 30:
```suggestion
):  Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const studentsCopy = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return studentsCopy.sort(
+        (a: Student, b: Student) => {
```
Apollonia5, line 37: maybe you should change naming for "a" and "b"

src/sortStudents.ts
Yevheniya-Deryka: yes)

src/sortStudents.ts
Yevheniya-Deryka: I thought if TS has already said a and b are Students, there's no need to repeat))

silchyn (changes requested):
`return studentsCopy.sort((a: Student, b: Student) => {` is repeated several times, so it would be nice if you restructure your code so that the switch-case is inside the compare function

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+            ? a[sortBy].localeCompare(b[sortBy])
+            : b[sortBy].localeCompare(a[sortBy]);
+        case SortType.AverageGrade:
+          return (order === 'asc')
+            ? getAvarageGrade(a[sortBy]) - getAvarageGrade(b[sortBy])
+            : getAvarageGrade(b[sortBy]) - getAvarageGrade(a[sortBy]);
+        default:
+          return (order === 'asc')
+            ? +a[sortBy] - (+b[sortBy])
+            : +b[sortBy] - (+a[sortBy]);
```
OleksiiBlezniuk, line 48: Use `Number` instead of `+`, it's more readable

OleksiiBlezniuk (approved):

PR #748, opened 2022-10-04
Solution

ArtemHulkoDev:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      studentsCopy.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+    } else {
+      studentsCopy.sort((a, b) => a[sortBy] - b[sortBy]);
+    }
+  } else if (order === 'desc') {
+    if (sortBy === 'grades') {
+      studentsCopy.sort((a, b) => (
+        (b[sortBy].reduce((sum, current) => sum + current) / b[sortBy].length)
+        - (a[sortBy].reduce((sum, current) => sum + current) / a[sortBy].length)
+      ));
```
Andrew-Skoryk, line 44: Would be better to use logic in separate function

Andrew-Skoryk (changes requested):
1. [CODE STYLE] - If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks

Anna-Tkachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+): number {
+  return order === 'asc'
+    ? (a[sortBy].reduce((sum: number, current: number) => sum + current)
+        / a[sortBy].length)
+      - (b[sortBy].reduce((sum: number, current: number) => sum + current)
+        / b[sortBy].length)
+    : (b[sortBy].reduce((sum: number, current: number) => sum + current)
+        / b[sortBy].length)
+      - (a[sortBy].reduce((sum: number, current: number) => sum + current)
+        / a[sortBy].length);
```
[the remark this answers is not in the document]
ArtemHulkoDev: done)

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+function getAverage(
+  a: Student,
+  b: Student,
+  sortBy: SortType,
+): number {
+  return (a[sortBy].reduce((sum: number, current: number) => sum + current)
+    / a[sortBy].length)
+    - (b[sortBy].reduce((sum: number, current: number) => sum + current)
+    / b[sortBy].length);
+}
```
Anna-Tkachenko, line 28: this function doesn't return some average value
it returns the result for the sort callback
rename function or refactor it's functionality
function name should match with functionality

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+        case SortType.Surname:
+          return stringSort(a[sortBy], b[sortBy], order);
 
+        case SortType.AverageGrade:
+          return gradesSort(a, b, order, sortBy);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+        default:
+          return numberSort(a[sortBy], b[sortBy], order);
```
Anna-Tkachenko, line 69: it's not default case, specify for what fields you want to sort array as numbers
and add default case with throwing an error with corresponding message

Anna-Tkachenko (changes requested):

src/sortStudents.ts
ArtemHulkoDev: renamed func

src/sortStudents.ts
ArtemHulkoDev: added specific fields

PR #749, opened 2022-10-04
src/sortStudents.ts

nikitaovch99:


src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+
+    case SortType.AverageGrade:
+      copyOfStudents.sort((studentA: object, studentB: object) => {
+        const studentAAverage = studentA[sortBy].reduce(
+          (previous: number, current: number) => previous + current,
+        ) / studentA[sortBy].length;
+
+        const studenBAverage = studentB[sortBy].reduce(
+          (previous: number, current: number) => previous + current,
+        ) / studentB[sortBy].length;
```
xiimiik, line 55: I think it's better to move this logic to a separate function

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): object[] {
+  const copyOfStudents: object[] = students.map(
+    (student: object): object => {
+      return { ...student };
+    },
+  );
```
xiimiik, line 28:
```suggestion
  const copyOfStudents= [...students];
```
in this situation it is enough to do so

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyOfStudents.sort((studentA: object, stundentB: object) => (
```
xiimiik, line 33:
```suggestion
      copyOfStudents.sort((studentA: Student, stundentB: Student) => (
```
we must type not just as an object, but as an interface. do it in the code below

xiimiik (changes requested):

src/sortStudents.ts
nikitaovch99: In the procces i thought I will change properties but eventually i realized that i don't  have to but forgot to change it, will fix.

src/sortStudents.ts
nikitaovch99: I understand but everytime I wrote that i had this problem 
image

src/sortStudents.ts
nikitaovch99: I think it's not that necessary here.

xiimiik: let it be

src/sortStudents.ts
xiimiik: it can be because you subtract: true - false, false - true, etc...
you need to convert that values into numbers

nikitaovch99: I changed it but still problem remain and even for a different cases, but anyway I changed all for Student so...
image

xiimiik (approved):

src/sortStudents.ts
nikitaovch99: fixed this

src/sortStudents.ts
OleksiiBlezniuk: I agree with @xiimiik, this `reduce` used twice, and helper function that will calculate average value could be useful in project, let's move it to separate function

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyOfStudents = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyOfStudents.sort((studentA: Student, stundentB: Student) => (
```
OleksiiBlezniuk, line 29: You can move this switch/case inside `sort` method, no need to write it every time 🙂

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
Good job!

PR #750, opened 2022-10-04
Task was solved

DenysKleimenov:


Anna-Borodaienko (commented):
I think it would be better to create special functions for sorting and use them in the main code

DenysKleimenov:
> I think it would be better to create special functions for sorting and use them in the main code
I don't think that it's necessary, I would have to create separate functions for different sortings that we do just once in our code, there is no point in this, we have only two different ways for our sortings according to conditions, so ternary operators do just fine + as it seems to me it could cause some problems understanding the code (if using the same function for numbers and booleans)

OleksiiBlezniuk:
> > I think it would be better to create special functions for sorting and use them in the main code
> 
> I don't think that it's necessary, I would have to create separate functions for different sortings that we do just once in our code, there is no point in this, we have only two different ways for our sortings according to conditions, so ternary operators do just fine + as it seems to me it could cause some problems understanding the code (if using the same function for numbers and booleans)

Agree with @DenysKleimenov, in this case it's not so necessary. Nothing bad will happens if we do it, but we can leave as it is now 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+            : studentB[sortBy].localeCompare(studentA[sortBy])),
+      );
+
+    case SortType.Married:
+      return sortedStudents.sort(
+        (studentA: Student, studentB: Student) => (
+          order === 'asc'
+            ? Number(studentA[sortBy]) - Number(studentB[sortBy])
+            : Number(studentB[sortBy]) - Number(studentA[sortBy])),
+      );
```
OleksiiBlezniuk, line 57: You've used same sorting for `Age`, so let's just combine this cases 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
 
+    case SortType.AverageGrade:
+      return sortedStudents.sort(
```
OleksiiBlezniuk, line 60: Also you can move this switch/case to sort method and avoid repeating of this method writing every time

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

src/sortStudents.ts
DenysKleimenov: Oh, yeah, I thought about this, but wasn't really sure :)

PR #752, opened 2022-10-04
solution was added

Maksym845:


Yeliseiev (approved):
Good job! I think your solution is more convenient than mine🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+      ));
+      break;
+
+    case SortType.Married:
+      sortedListOfStudents.sort((person: Student, prevPerson: Student) => (
+        order === 'asc'
+          ? Number(person[sortBy]) - Number(prevPerson[sortBy])
+          : Number(prevPerson[sortBy]) - Number(person[sortBy])
+      ));
+      break;
```
OleksiiBlezniuk, line 51: It's a same logic, let's combine cases

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedListOfStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      sortedListOfStudents.sort((person: Student, prevPerson: Student) => (
```
OleksiiBlezniuk, line 30: We can move switch/case inside `sort` method and just return needed comparing result, why duplicate this line of code every time? 🙂

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function getAverageValue(numbers: number[]): number {
+  return numbers.reduce((prev, curr) => prev + curr) / numbers.length;
```
sergii-nosachenko, line 21:
```suggestion
  return numbers.reduce((prev, curr) => prev + curr, 0) / numbers.length;
```
Add starting value every time, it's safer and bug proof.

sergii-nosachenko (approved):
Good job 👍
Read the comment.

PR #753, opened 2022-10-04
Done

iPanda177:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? copy.sort((firstStudent, secondStudent) => {
+          return firstStudent[sortBy].localeCompare(secondStudent[sortBy]);
+        })
+        : copy.sort((firstStudent, secondStudent) => {
+          return secondStudent[sortBy].localeCompare(firstStudent[sortBy]);
+        });
```
iam-datsenko, line 41: u don't need to return 3 times, write 1 return before 'copy.sort....'
and u can also sort 1 time and use a ternary operator for other options

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? copy.sort((firstStudent, secondStudent) => {
+          return +(firstStudent[sortBy]) - +(secondStudent[sortBy]);
+        })
+        : copy.sort((firstStudent, secondStudent) => {
+          return +(secondStudent[sortBy]) - +(firstStudent[sortBy]);
+        });
```
iam-datsenko, line 51: same as before, do it for all ur cases

iam-datsenko (changes requested):

silchyn (changes requested):
`copy.sort((firstStudent, secondStudent) => (` is repeated many times, so it would be fantastic if you restructure your code so that the switch-case is inside the compare function

silchyn (changes requested):
`return copy.sort((firstStudent, secondStudent) => (` is repeated several times, so it would be nice if you restructure your code so that the switch-case is in the compare function

PR #756, opened 2022-10-04
solution

vmamchur:


jojnajbude (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      ));
+      break;
+
+    case SortType.Married:
+      sortedStudents.sort((personA: Student, personB: Student) => (
+        order === 'asc'
+          ? Number(personA[sortBy]) - Number(personB[sortBy])
+          : Number(personB[sortBy]) - Number(personA[sortBy])
+      ));
+      break;
```
OleksiiBlezniuk, line 57: In both cases we using same sorting, let's combine this cases

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+
+    case SortType.AverageGrade:
+      sortedStudents.sort((personA: Student, personB: Student) => (
```
OleksiiBlezniuk, line 60: Also you can move switch/case inside sort callback instead of writing this method every time

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #758, opened 2022-10-04
sortStudent

Arsen-Shpak:


Dikey945:
Nice! 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
 export enum SortType {
   // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
+function AverageGrades(grade:number[]):number {
```
Anna-Tkachenko, line 20: function name should be an action
AverageGrades it's not an action

Anna-Tkachenko (approved):

PR #759, opened 2022-10-04
Solution

OleksandrGlulhovskiy:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
-  // write your function
+    case SortType.AverageGrade:
+      return studentsCopy.sort((prev: Student, curr: Student) => (
+        order === 'asc'
+          ? getAverage(prev[sortBy]) - getAverage(curr[sortBy])
+          : getAverage(curr[sortBy]) - getAverage(prev[sortBy])
+      ));
+    default:
+      return studentsCopy;
+  }
```
grafit-off, line 58: I prefer to add a `switch case` into the `sort` method, so we don't need to copy `sort` all the time, but this is still good.

grafit-off (approved):
GJ 🔥

OleksiiBlezniuk (approved):

PR #761, opened 2022-10-04
Solution

imaksweb:


artem-musii (approved):
Cool solution

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+    const averageGrade: number = student.grades
+      .reduce((total, current) => total + current) / student.grades.length;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    return {
+      ...student,
+      averageGrade,
+    };
+  });
```
OleksiiBlezniuk, line 32: But why do we need to map this array if sortBy will be name? Or any other except averageGrade?

Let's just calculate this value inside AverageGrade case

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+        ? (a, b): number => Number(a[sortBy]) - Number(b[sortBy])
+        : (a, b): number => Number(b[sortBy]) - Number(a[sortBy]);
+      break;
+
+    case SortType.AverageGrade:
+      sortingCallback = (a, b): number => {
+        const averageGradeA = a[sortBy]
+          .reduce((total, current) => total + current) / a[sortBy].length;
+        const averageGradeB = b[sortBy]
+          .reduce((total, current) => total + current) / b[sortBy].length;
```
OleksiiBlezniuk, line 47: Ok, but better to move it to separate helper function

OleksiiBlezniuk (approved):

PR #771, opened 2022-10-04
solution

myroslav-diiak:


anastasiia-nahaiska (approved):
GJ))

silchyn (changes requested):
Good points above

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrades(grades: number[]): number {
+  return grades.reduce((prev, curr) => prev + curr) / grades.length;
+}
```
OleksiiBlezniuk, line 24: As for me it's more logical to have universal logic in helper functions in order to reuse it with other values, so better to use more generic names in helpers (if it has common logic). For example this helper can be used to calculate average value from any number array, not only for grades

OleksiiBlezniuk, line 24:
```suggestion
function getAverageValue(numbers: number[]): number {
  return numbers.reduce((prev, curr) => prev + curr) / numbers.length;
}
```

OleksiiBlezniuk (approved):

PR #774, opened 2022-10-04
Solution

tatyana22:


src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+            nextPerson: Student) => person[sortBy] - nextPerson[sortBy]);
+      }
+
+      return sortedStudents
+        .sort((person: Student,
+          nextPerson: Student) => nextPerson[sortBy] - person[sortBy]);
+
+    case SortType.AverageGrade:
+      sortedStudents.sort((person: Student, nextPerson: Student) => {
+        const prevPersonAverage = person[sortBy]
```
volodymyrklymkiv, line 60: u can make separate function for this because u use sam code more then one time

src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
 
 export interface Student {
   // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
+  middle: number;
```
volodymyrklymkiv, line 9: i cant find in code middle, i think we didnt even use this

volodymyrklymkiv (approved):
great, just leave some comments

src/sortStudents.ts
Anna-Tkachenko: where did you find such property?

src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+    case SortType.Surname:
+      if (order === 'asc') {
+        return sortedStudents
+          .sort((person: Student, nextPerson: Student) => (person[sortBy])
+            .localeCompare(nextPerson[sortBy]));
+      }
+
+      return sortedStudents
+        .sort((person: Student, nextPerson: Student) => (nextPerson[sortBy])
+          .localeCompare(person[sortBy]));
```
Anna-Tkachenko, line 44: for now you write the same code too many times
let's move if statement inside the sort method

src/sortStudents.ts
Anna-Tkachenko: agree, don't repeat yourself
create separate function for calculating average grade

Anna-Tkachenko (changes requested):

src/sortStudents.ts
tatyana22: it was previous solution, which dont work) ok, delete

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok!

tatyana22:
Ok

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+// function sortLiteral((a, b) => a.localeCompare(b);
```
OleksiiBlezniuk, line 23: remove commented code

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+// function sortNumbers((a, b) => a - b;
+
+function findAverageGrade(grades: number[]): number {
+  return grades.reduce((sum: number, x: number) => sum + x, 0) / grades.length;
+}
```
OleksiiBlezniuk, line 29: Ok now, but it's more logical to create helper functions that can be reused for other values. So it's more logical to create more generic names for it instead of specific. For example this function could be named as getAverage and accept array of generic `numbers` instead of specific `grades` to be reusable for other arrays too, not only for grades 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortedStudents.sort((person: Student, nextPerson: Student) => {
```
OleksiiBlezniuk, line 41: Or you could move switch/case inside `sort` method and avoid writing this method every time

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+          return findAverageGrade(person[sortBy])
+            - findAverageGrade(nextPerson[sortBy]);
+        }
+
+        return findAverageGrade(nextPerson[sortBy])
+          - findAverageGrade(person[sortBy]);
+      });
+
+      break;
+
```
OleksiiBlezniuk, line 71: Why do you need break here? You can just return sorted array

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+    default:
+
+      break;
```
OleksiiBlezniuk, line 74: Same here

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok

src/sortStudents.ts
tatyana22: ok

Denchepornyuk (approved):

PR #775, opened 2022-10-04
Solution

korniienkokarina:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  [...students]: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] | string {
```
Rina07, line 24:
```suggestion
  order: SortOrder,
): Student[]  {
```

Rina07: Looks like it returns just array, right? :)

Rina07 (approved):
GJ :star: Interesting decision with copying students array :)

src/sortStudents.ts
Rina07: Looks like it returns just array, right? :)

Andr1yK (approved):

PR #776, opened 2022-10-04
Solution

vallerii:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  ), 0) / grades.length;
+}
+
+export function sortStudents(students, sortBy, order): Student[] {
+  const copyOfStudents = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyOfStudents.sort((studentA: Student, studentB: Student): number => {
```
silchyn, line 33: This line repeats several times, better to write it one time and use switch-case inside the compare function

silchyn (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(grades: number[]): number {
```
naziamoff, line 21: normally function name starts with a verb
```suggestion
function getAverageGrade(grades: number[]): number {
```

naziamoff (approved):
GJ 🔥

PR #777, opened 2022-10-04
Solution

AniaHna:


AntonVaida (approved):
looks good

naziamoff (changes requested):
You did not fix Ivan's comment. You repeat the same logic in each case of the switch.

Try to implement switch inside the sort

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(arr: number[]): number {
```
Anna-Tkachenko, line 20: don't use data type names in variable names
create more meaningful variable name instead of `arr`

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAverage(firstStudent[sortBy])
+            - getAverage(secondStudent[sortBy])
+          : getAverage(secondStudent[sortBy])
+            - getAverage(firstStudent[sortBy]);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default: throw new Error('Wrong');
```
Anna-Tkachenko, line 52: specify understandable text messages for the errors, what's wrong here?

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #778, opened 2022-10-04
solution of sort students task

Dikey945:


src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+      ));
+      break;
+
+    case SortType.Married:
+      sortedStudents.sort((studentA: Student, studentB: Student) => (
+        order === 'asc'
+          ? Number(studentA[sortBy]) - Number(studentB[sortBy])
+          : Number(studentB[sortBy]) - Number(studentA[sortBy])
+      ));
+      break;
```
artem-musii, line 60: You can combine these two methods, they have the same logic

artem-musii (approved):
GJ! Just a tip: instead of repeating sort method you can just move switch/case inside it

src/sortStudents.ts
OleksiiBlezniuk, line 60: Combine this cases, no need to duplicate logic

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+
+    case SortType.AverageGrade:
+      sortedStudents.sort((studentA: Student, studentB: Student) => (
```
OleksiiBlezniuk, line 63: Also you can move switch/case inside `sort` method, no need to write it every time 🙂

OleksiiBlezniuk (changes requested):

Andr1yK (approved):

PR #780, opened 2022-10-04
solution

poznianski:


src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
   // write your function
+  return [...students]
+    .sort((
+      prevStudent: Student,
+      currStudent: Student,
+    ): number => {
+      switch (sortBy) {
+        case SortType.Name:
+        case SortType.Surname:
+          return stringCompare(prevStudent[sortBy], currStudent[sortBy], order);
```
anastasiia-nahaiska, line 58: Good solution)

anastasiia-nahaiska (approved):
GJ))

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+  return curr - prev;
+}
+
+function stringCompare(prev: string, curr: string, order: string): number {
+  if (order === 'asc') {
+    return prev.localeCompare(curr);
+  }
 
+  return curr.localeCompare(prev);
+}
```
OleksiiBlezniuk, line 42: Nice 👍

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+            getAverageGrades(currStudent[sortBy]),
+            order,
+          );
+
+        default:
+          return numberCompare(
+            Number(prevStudent[sortBy]),
+            Number(currStudent[sortBy]),
+            order,
+          );
```
OleksiiBlezniuk, line 72: Ok for this task, but I think we may return something like 0 to avoid sorting in case if we'll add other `sortBy` type and didn't change logic of this function

OleksiiBlezniuk (approved):

PR #781, opened 2022-10-04
Students sort

Yuriy-Komolov:


oleksandr-korobko (approved):
Good job.

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  [...students]: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  switch (sortBy) {
```
Andr1yK, line 29: Let's use one sort method.. and one switch case inside😉

Andr1yK (changes requested):

PR #783, opened 2022-10-05
Solution

avkovlit:


imaksweb (approved):
GJ 👍

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(studentGrades: number[]): number {
+  return studentGrades.reduce((accum: number, current: number) => (
+    accum + current
+  ), 0) / studentGrades.length;
+}
```
OleksiiBlezniuk, line 24: In that case better to create more generic name for incoming array. I mean this helper function can be used for other values too, not only for studentsGrades

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(studentGrades: number[]): number {
```
OleksiiBlezniuk, line 20:
```suggestion
function getAverage(numbers: number[]): number {
```

OleksiiBlezniuk, line 20:
```suggestion
function getAverage(numericalValues: number[]): number {
```

OleksiiBlezniuk (approved):

PR #784, opened 2022-10-05
Solution

VovaDzhuryk:


Andrew-Skoryk (changes requested):
Put all switches/cases inside the sort function. Also, you can use `orderMultiplier` once, when you return the result from the sort function.

PR #786, opened 2022-10-05
result

AndriyZhabynets:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAvgGrade(grades: number[]): number {
+  const sumOfGrades: number = grades
+    .reduce((sum: number, current: number) => sum + current);
```
glugovsky, line 22: Not necessary for tests, but it would be nice to set initial zero value for reducer
```suggestion
    .reduce((sum: number, current: number) => sum + current, 0);
```

glugovsky (approved):

Denchepornyuk (approved):

PR #793, opened 2022-10-09
Solution

DimaFlacko:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+          : Number(curr[sortBy]) - Number(prev[sortBy]);
+
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? prev[sortBy]
+            .reduce((a, b) => a + b) / prev[sortBy].length - curr[sortBy]
+            .reduce((a, b) => a + b) / curr[sortBy].length
+          : curr[sortBy]
+            .reduce((a, b) => a + b) / curr[sortBy].length - prev[sortBy]
+            .reduce((a, b) => a + b) / prev[sortBy].length;
```
Andrew-Skoryk, line 50: Would be better to use this logic in a separate function

Andrew-Skoryk (approved):
GJ!

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return prev[sortBy].localeCompare(curr[sortBy]);
+
+      case SortType.Age:
+        return order === 'asc'
+          ? prev[sortBy] - curr[sortBy]
+          : curr[sortBy] - prev[sortBy];
+
+      case SortType.Married:
```
Denchepornyuk, line 38: you can combine these cases into one

src/sortStudents.ts
Denchepornyuk: Agree, let's create a function to repeat code

Denchepornyuk (changes requested):

Andr1yK (approved):

PR #796, opened 2022-10-12
Students sorted

zhenija:


src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  AverageGrade
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverage(gradesArray: number[]): number {
+  const average = gradesArray.reduce((sum, x) => sum + x,
+    0) / gradesArray.length;
```
Anastasiia-Svintsova, line 22:
```suggestion
  const average = gradesArray.reduce((sum, x) => sum + x, 0) / gradesArray.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  Age,
+  Married,
+  AverageGrade
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverage(gradesArray: number[]): number {
```
Anastasiia-Svintsova, line 20: It is better to give clear naming. Average what?
```suggestion
function calculateAverageGrade(gradesArray: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  return sortedArray.sort((
+    a: Student,
+    b: Student,
+  ): number => {
+    switch (sortBy) {
+      case SortType.Name:
+        return (order === 'asc')
+          ? a.name.localeCompare(b.name)
+          : b.name.localeCompare(a.name);
+
```
Anastasiia-Svintsova, line 43: look, you check `order === 'asc'` in all cases. Think about how to do it once at the beginning of the function

Anastasiia-Svintsova (changes requested):
Fix some comments

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverageGrade(gradesArray: number[]): number {
+  const average = gradesArray.reduce((sum, x) => sum + x,
+    0) / gradesArray.length;
```
qoa11a, line 22:
```suggestion
  const average = gradesArray
    .reduce((sum, x) => sum + x, 0) / gradesArray.length;
```

src/sortStudents.ts
qoa11a: It's okay this way. The other possible solution is to check if the order is 'asc' at the top level.
For example, you could write

image

qoa11a (approved):
Good job! Few comments, but they are rather recommendations than mistakes.

PR #800, opened 2022-10-14
solution

bradleymatjie:


kamal-tufekcic (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
```
ihor-jpeg, line 21: why not enum?

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+    case SortType.AverageGrade:
+      studentsCopy.sort((stu1: Student, stu2: Student): number => {
+        return order === 'asc'
+          ? averageGrades(stu1[sortBy]) - averageGrades(stu2[sortBy])
+          : averageGrades(stu2[sortBy]) - averageGrades(stu1[sortBy]);
+      });
+      break;
+
+    default:
+      return [];
```
ihor-jpeg, line 62: maybe by default you should return the original array?

ihor-jpeg (changes requested):

VinnieJ-2k20 (approved):

PR #803, opened 2022-10-18
Solution

MeshackDev:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    }
+  }
 
+  if (sortBy === SortType.AverageGrade) {
+    if (order === 'asc') {
+      result.sort((a, b) => getAverage(a) - getAverage(b));
+    } else {
+      result.sort((a, b) => getAverage(b) - getAverage(a));
+    }
+  }
```
kamal-tufekcic, line 69: Why not switch?

kamal-tufekcic (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const result: Student[] = [...students];
```
ihor-jpeg, line 24: here type will be inferred

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    if (order === 'asc') {
+      result.sort((a, b) => +a.married - +b.married);
+    } else {
+      result.sort((a, b) => +b.married - +a.married);
+    }
+  }
 
+  if (sortBy === SortType.AverageGrade) {
+    if (order === 'asc') {
+      result.sort((a, b) => getAverage(a) - getAverage(b));
```
ihor-jpeg, line 65: switch

ihor-jpeg (changes requested):

src/sortStudents.ts
MeshackDev: Noted, changes made.

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+        ? result.sort((a, b) => a.age - b.age)
+        : result.sort((a, b) => b.age - a.age);
+    case SortType.Married:
+      return order === 'asc'
+        ? result.sort((a, b) => +a.married - +b.married)
+        : result.sort((a, b) => +b.married - +a.married);
+    default:
+      return order === 'asc'
+        ? result.sort((a, b) => getAverage(a) - getAverage(b))
+        : result.sort((a, b) => getAverage(b) - getAverage(a));
```
VinnieJ-2k20, line 51: It should not be default case, it needs own case.
Better to return untouched `result` as default.

VinnieJ-2k20 (changes requested):
One thing to fix.

src/sortStudents.ts
MeshackDev: Got it, changes made.

VinnieJ-2k20 (approved):

PR #806, opened 2022-10-18
First commit

Bibicutu:


src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  for (let i = 0; i < sortedStudents.length; i += 1) {
+    sortedStudents[i].averageGrade = sortedStudents[i].grades.reduce(
+      (sum: number, grade: number) => {
+        return sum + grade;
+      }, 0,
+    ) / sortedStudents[i].grades.length;
+  }
```
kamal-tufekcic, line 38: Do you really need a for loop here?

kamal-tufekcic (approved):

src/sortStudents.ts
ihor-jpeg, line 38: just reduce should be enough

src/sortStudents.ts
```diff
@@ -1,16 +1,88 @@
...
+
+  switch (sortBy) {
+    case 'NAME':
```
ihor-jpeg, line 41: use the enum

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #809, opened 2022-10-21
Solution

bokorzita:


amitpatiljc (changes requested):
Instead of if statements use switch case for `sortBy` case

amitpatiljc (approved):
Nicely Done!

PR #816, opened 2022-11-04
Solution

anastasia-babko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getGrades(grade: number[]): number {
+  return grade.reduce((a, b) => (a + b), 0) / grade.length;
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] | 0 {
```
Anastasiia-Svintsova, line 26: You can return by default just studentsCopy to avoid this 0 in the return type
```suggestion
export function sortStudents(
  students: Student[], 
  sortBy: SortType, 
  order: SortOrder,
): Student[] {
```

Anastasiia-Svintsova (changes requested):
Good! Just fix one little comment

SanyaBratashchuk (approved):

PR #824, opened 2022-11-09
made sort students with TS

Hlazman:


src/sortStudents.ts
```diff
@@ -1,16 +1,89 @@
...
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      result.sort((a, b) => {
+        if (order === 'asc') {
+          return a.name.localeCompare(b.name);
+        }
+
+        return b.name.localeCompare(a.name);
```
vlad-demchuk, line 35: use ternary operator here. it'll be better for readability

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #828, opened 2022-11-09
Solution

taraselnik:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    case SortType.Surname:
+      toSortStudents.sort(
+        (currentStudent: Student,
+          nextStudent: Student): number => {
+          return (order === 'asc')
+            ? currentStudent[sortBy].localeCompare(nextStudent[sortBy])
+            : nextStudent[sortBy].localeCompare(currentStudent[sortBy]);
+        },
+      );
+      break;
```
DanilSerzh, line 42: break isn't necessary, because anyway, and maybe you should add throw error in default case

DanilSerzh (changes requested):

src/sortStudents.ts
taraselnik: thx) but I like how it looks with with breaks after each case-block;) ok, ill add errror

illiarerun, line 42: no need for break here, you can just return your result

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    case SortType.Married:
+      toSortStudents.sort(
+        (currentStudent: Student,
+          nextStudent: Student): number => {
+          return (order === 'asc')
+            ? Number(currentStudent[sortBy]) - Number(nextStudent[sortBy])
+            : Number(nextStudent[sortBy]) - Number(currentStudent[sortBy]);
+        },
+      );
+      break;
```
illiarerun, line 54: same here

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    case SortType.Name:
+    case SortType.Surname:
+      toSortStudents.sort(
+        (currentStudent: Student,
+          nextStudent: Student): number => {
+          return (order === 'asc')
+            ? currentStudent[sortBy].localeCompare(nextStudent[sortBy])
+            : nextStudent[sortBy].localeCompare(currentStudent[sortBy]);
+        },
+      );
```
illiarerun, line 41: fix you formatting everywhere
```suggestion
      toSortStudents.sort(
        (currentStudent: Student, nextStudent: Student): number => {
          return (order === 'asc')
            ? currentStudent[sortBy].localeCompare(nextStudent[sortBy])
            : nextStudent[sortBy].localeCompare(currentStudent[sortBy]);
        },
      );
```

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const arrAvg = (arr:number[]): number => arr
+  .reduce((a, b) => a + b, 0) / arr.length;
```
illiarerun, line 22:
```suggestion
const arrAvg = (arr: number[]): number => (
  arr.reduce((a, b) => a + b, 0) / arr.length
);
```

illiarerun (changes requested):

illiarerun (approved):

PR #829, opened 2022-11-10
Solution

Esceype:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calcAvgGrade(arrOfGrades: number[]): number {
+  return arrOfGrades.reduce((x: number, y:number) => {
+    return x + y;
+  }) / arrOfGrades.length;
```
Anna-Shapovalova, line 24:
```suggestion
  return arrOfGrades.reduce((x: number, y:number) => x + y, 0) / arrOfGrades.length;
```
The result in the callback is too short - U can return it just like this.  Also, I think it`s a good idea to set the initial value of the accumulator)

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
```
Anna-Shapovalova, line 29:
```suggestion
export function sortStudents(
  students: Student[], 
  sortBy: SortType, 
  order: SortOrder,
): Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+      });
+
+    case SortType.Age:
+    case SortType.Married:
+      return copyOfSt.sort((x: Student, y: Student) => {
+        return order === 'asc'
+          ? x[sortBy] - y[sortBy]
+          : y[sortBy] - x[sortBy];
```
Anna-Shapovalova, line 46:
```suggestion
          ? +x[sortBy] - +y[sortBy]
          : +y[sortBy] - +x[sortBy];
```
How does it work with boolean values?

Anna-Shapovalova (approved):
Nice code) 🚀  Well done) Pay attention to some comments)

src/sortStudents.ts
mykola-koval: agree

src/sortStudents.ts
mykola-koval: agree

mykola-koval (changes requested):
make fixes after mentor's review

src/sortStudents.ts
Esceype: false == 0
true == 1

mykola-koval (approved):

PR #831, opened 2022-11-10
Solution

andriy-shymkiv:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+          ? pupil1.age - pupil2.age
+          : pupil2.age - pupil1.age;
+      });
+
+    case SortType.Married:
+      return studentsCopy.sort((pupil1: Student, pupil2: Student) => {
+        return order === 'asc'
+          ? +pupil1.married - +pupil2.married
+          : +pupil2.married - +pupil1.married;
+      });
```
karyna-salnykova, line 61: I think you can use here
`case 1:
case 2:
  //code`

to make your code shorter and unite statements that do the same

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+          ? pupil1.name.localeCompare(pupil2.name)
+          : pupil2.name.localeCompare(pupil1.name);
+      });
+
+    case SortType.Surname:
+      return studentsCopy.sort((pupil1: Student, pupil2: Student) => {
+        return order === 'asc'
+          ? pupil1.surname.localeCompare(pupil2.surname)
+          : pupil2.surname.localeCompare(pupil1.surname);
+      });
```
karyna-salnykova, line 47: The same advice as in SortType.Age and SortType.Married part

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+  grades: Array<number>;
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
```
karyna-salnykova, line 15: I think you should write which value of Student should be in every part. Don't know how to explain what I mean, but it's like:

`Name = 'name'`

karyna-salnykova (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    .reduce((acc: number, mark: number) => acc + mark, 0) / gradesArray.length;
+}
+
+export function sortStudents(
+  students: Array<Student>,
+  sortBy: SortType,
+  order: SortOrder,
+)
+
+  : Array<Student> {
```
illiarerun, line 31: formatting
```suggestion
)  : Array<Student> {
```

illiarerun (approved):

PR #832, opened 2022-11-10
Solution

karyna-salnykova:


Stepico (approved):
Wow, because I was thinking, like, what should I do to follow the DRY rule, your solution is awesome

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+
+    case SortType.AverageGrade:
+      return studentsCopy.sort((first: Student, second: Student) => {
+        if (order === 'desc') {
+          return getAverageGrade(second[sortBy])
+            - getAverageGrade(first[sortBy]);
+        }
+
+        return getAverageGrade(first[sortBy])
+            - getAverageGrade(second[sortBy]);
```
illiarerun, line 64:
```suggestion
        return getAverageGrade(first[sortBy])
          - getAverageGrade(second[sortBy]);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+      });
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      throw new Error('Change your input');
```
illiarerun, line 68: change what exactly, please specify

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
+  return grades
+    .reduce((sum: number, current: number) => sum + current, 0)
+    / grades.length;
```
illiarerun, line 24: would be easier to read this way
```suggestion
  return grades.reduce((sum: number, current: number) => (
    sum + current
  ), 0)/ grades.length;
```

illiarerun (approved):

PR #833, opened 2022-11-10
Task was Solved

Stepico:


src/sortStudents.ts
```diff
@@ -1,16 +1,110 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+      if (order === 'asc') {
+        sortedStudents.sort((a, b) => {
+          return a.name.localeCompare(b.name);
+        });
+      }
+
+      if (order === 'desc') {
```
karyna-salnykova, line 35: There's no need to use another if-statement. You have two values: if one is not valid, the other should be.

src/sortStudents.ts
```diff
@@ -1,16 +1,110 @@
...
+        sortedStudents.sort((a, b) => {
+          return b.name.localeCompare(a.name);
+        });
+      }
+
+      break;
```
karyna-salnykova, line 41: Why you use `break` here?

src/sortStudents.ts
```diff
@@ -1,16 +1,110 @@
...
+        sortedStudents.sort((a, b) => {
+          return +a.married - +b.married;
+        });
+      }
+
+      if (order === 'desc') {
+        sortedStudents.sort((a, b) => {
+          return +b.married - +a.married;
+        });
+      }
```
karyna-salnykova, line 84: I think you can use here:
`case 1:` 
`case 2:`
`//code`
to make your code shorter and unite statements that do the same

src/sortStudents.ts
```diff
@@ -1,16 +1,110 @@
...
+        sortedStudents.sort((a, b) => {
+          return b.grades.reduce((prev, cur) => prev + cur, 0) / b.grades.length
+            - a.grades.reduce((prev, cur) => prev + cur, 0) / a.grades.length;
+        });
+      }
+
+      break;
 
+    default:
+      throw new Error('Why? I thought TS will handle it');
```
karyna-salnykova, line 106: I think here you should write what the problem exactly is, to make sure the user understands what he should change

karyna-salnykova (changes requested):

src/sortStudents.ts
Stepico: Absolutely agree

src/sortStudents.ts
Stepico: That's where it becomes tricky for me, because of this connection between JS and TS, I don't really understand it, because I thought that TS does its job before JS, so we pointed out that we have only a specific range of values, so default is redundant

src/sortStudents.ts
Stepico: Because I didn't use return in my cases, maybe I should have

src/sortStudents.ts
```diff
@@ -1,16 +1,110 @@
...
+  grades: number[],
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
```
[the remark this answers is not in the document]
Stepico: I want to hear the mentor's thoughts about it because I don't understand it, I thought that when we have such writing, it means that we have indexes as keys, and when we have something like you wrote, then our fields will be the keys, but it works and I don't know why and how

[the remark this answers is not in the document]
illiarerun: No, in such case we would have our `fields` as `keys`, and `indeces` as values

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+)
+  : object {
```
illiarerun, line 24: formatting
```suggestion
export function sortStudents(
  students: Student[],
  sortBy: SortType,
  order: SortOrder,
): object {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+          return +a[sortBy] - +b[sortBy];
+        }
+
+        return +b[sortBy] - +a[sortBy];
+      });
+
+    case SortType.AverageGrade:
+      return sortedStudents.sort((a, b) => {
+        if (order === 'asc') {
+          return a.grades.reduce((prev, cur) => prev + cur, 0) / a.grades.length
```
illiarerun, line 51: DRY: create a method for reduce call + division

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+        }
 
+        return b.grades.reduce((prev, cur) => prev + cur, 0) / b.grades.length
+        - a.grades.reduce((prev, cur) => prev + cur, 0) / a.grades.length;
+      });
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      throw new Error('Please, input the correct field');
```
illiarerun, line 60: what field exactly?

illiarerun (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+function calcAvgGrade(grades: number[]): number {
+  return grades.reduce((prev, cur) => prev + cur, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+) : object {
```
denys-danyliuk, line 27: let's specify what this function actually returns

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortedStudents.sort((a, b) => {
+        if (order === 'asc') {
+          return a[sortBy].localeCompare(b[sortBy]);
+        }
+
+        return b.surname.localeCompare(a.surname);
+      });
```
denys-danyliuk, line 40: use ternary

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+) : object {
+  const sortedStudents: Student[] = [...students];
+  const values = Object.values(SortType);
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortedStudents.sort((a, b) => {
```
denys-danyliuk, line 34: you can do this before switch statement to avoid doing it in every case

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #840, opened 2022-11-10
sort students

Svitlana-Yudina:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+export function getAverageGrade(grade: number[]): number {
+  return grade
+    .reduce((prev: number, el: number) => (prev + el), 0) / grade.length;
 }
 
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
```
Anna-Shapovalova, line 25:
```suggestion
  students: Student[],
  sortBy: SortType, 
  order: SortOrder,
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+): object[] {
```
Anna-Shapovalova, line 26:
```suggestion
): Student[] {
```
U return the same objects, but sorted)

Anna-Shapovalova (approved):
Well done!))) 🚀

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const stArr: Student[] = students.map((el: Student) => ({ ...el }));
+
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
```
Anna-Shapovalova, line 31:
```suggestion
    case SortType.Name
    case SortType.Surname:
```
Why don`t U use enum keys?)

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      return (order === 'asc')
+        ? stArr
+          .sort((a: Student, b: Student) => (
+            a[sortBy].localeCompare(b[sortBy])))
+        : stArr
+          .sort((a: Student, b: Student) => (
+            b[sortBy].localeCompare(a[sortBy])));
+
+    case 'age':
+    case 'married':
```
Anna-Shapovalova, line 41:
```suggestion
    case SortType.Age:
    case SortType.Married:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      return (order === 'asc')
+        ? stArr
+          .sort((a: Student, b: Student) => (
+            Number(a[sortBy]) - Number(b[sortBy])))
+        : stArr
+          .sort((a: Student, b: Student) => (
+            Number(b[sortBy]) - Number(a[sortBy])));
 
+    case 'grades':
```
Anna-Shapovalova, line 50:
```suggestion
    case SortType.AverageGrade:
```

Anna-Shapovalova (approved):

src/sortStudents.ts
Svitlana-Yudina: You're right) Thanks))

src/sortStudents.ts
Svitlana-Yudina: I don't think it is necessary, but maybe you're right))

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
 }
 
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const stArr: Student[] = students.map((el: Student) => ({ ...el }));
```
denys-danyliuk, line 29: use spread operator

src/sortStudents.ts
denys-danyliuk: it is actually)
it's one of the main point why we create enum in this task

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
+      return stArr.sort((a: Student, b: Student) => {
```
denys-danyliuk, line 34: you can do this once before switch statement to avoid doing it in every case

denys-danyliuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+export type SortOrder = 'asc' | 'desc';
+
+export function getAverageGrade(grade: number[]): number {
```
denys-danyliuk, line 19:
```suggestion
export function getAverageGrade(grades: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const stArr: Student[] = [...students];
+
+  stArr.sort((a: Student, b: Student) => {
```
denys-danyliuk, line 31:
```suggestion
 [...students].sort((a: Student, b: Student) => {
```
you can do copy here instead of creating variable

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+          ? Number(a[sortBy]) - Number(b[sortBy])
+          : Number(b[sortBy]) - Number(a[sortBy]);
+
+      case SortType.AverageGrade:
+        return (order === 'asc')
+          ? getAverageGrade(a[sortBy]) - getAverageGrade(b[sortBy])
+          : getAverageGrade(b[sortBy]) - getAverageGrade(a[sortBy]);
 
+      default:
+        throw new Error('invalid data!');
```
denys-danyliuk, line 51: error message should be more representative

denys-danyliuk (approved):

PR #843, opened 2022-11-10
solution

Anna-Shapovalova:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  const studentsCopy: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return studentsCopy.sort(
+        (currentStudent: Student, nextStudent: Student) => {
+          return order === 'asc'
+            ? currentStudent[sortBy].localeCompare(nextStudent[sortBy])
+            : nextStudent[sortBy].localeCompare(currentStudent[sortBy]);
```
Svitlana-Yudina, line 39: 🔥 It's an awesome idea to use ternary operator inside the sort method)))

Anna-Shapovalova: Your school) I know, U like to use them, so I`m starting to do this too)

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(array: number[]): number {
```
denys-danyliuk, line 20:
```suggestion
function getAverageGrade(grades: number[]): number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return studentsCopy.sort(
+        (currentStudent: Student, nextStudent: Student) => {
```
denys-danyliuk, line 36: you can do this once before switch statement to avoid doing it in every case

denys-danyliuk (approved):

PR #846, opened 2022-11-10
Solution

referee1988:


andrii-bassarab (approved):
GJ!

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const cloneStudents: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return cloneStudents.sort((st1:Student, st2:Student) => {
```
denys-danyliuk, line 35: you can do this only once before switch statement instead of writing it in every case

denys-danyliuk (approved):

PR #848, opened 2022-11-10
solution

oksenyuk-valentyna:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+function getAverage(grades: number[]): number {
+  return grades.reduce((acc, grade) => acc + grade, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student,
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsArr = Object.values(students);
```
Anna-Shapovalova, line 29: Maybe in this case will be better to use a spread operator? If I'm not wrong, 'Object.values' doesn't copy an array of objects...

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+      case SortType.Age:
+        return (order === 'asc')
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
+
+      case SortType.Married:
+        return (order === 'asc')
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
```
Anna-Shapovalova, line 47: These two are the same - why not combine them? Also, how does it work with the boolean values?

Anna-Shapovalova (approved):
Nice code))) 👍  Left some comments)

src/sortStudents.ts
kryvets-vadym: agree, the spread operator is more appropriate for this case

src/sortStudents.ts
kryvets-vadym: agree, maybe the boolean can be converted to some other type and it becomes easy to sort

kryvets-vadym (approved):
Good job, just need to make a few fixes

mykola-koval (approved):

PR #849, opened 2022-11-10
Solution

yaro-bey:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentCopy: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return studentCopy.sort(
+        (firstStudent: Student, secondStudent: Student) => {
```
Anna-Shapovalova, line 36: U can do this once before the switch statement to avoid repeating - I got the same advice from the mentor)

Anna-Shapovalova (approved):
Well done!) 🚀

src/sortStudents.ts
illiarerun: agree

illiarerun (approved):

PR #864, opened 2022-11-13
Solution

alpopkov:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
+  const averageValue = (
+    grades: number[],
+  ): number => grades
+    .reduce((sum: number, num: number) => sum + num, 0) / grades.length;
```
karyna-salnykova, line 28: Maybe it will be better to write one more function for average grades?

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+        : studentsCopy.sort((a, b) => +b[sortBy] - +a[sortBy]);
+    case SortType.AverageGrade:
+      return (order === 'asc')
+        ? studentsCopy.sort(
+          (a, b) => averageValue(a.grades) - averageValue(b.grades),
+        )
+        : studentsCopy.sort(
+          (a, b) => averageValue(b.grades) - averageValue(a.grades),
+        );
+    default:
```
karyna-salnykova, line 49: You need to write something here

karyna-salnykova (changes requested):
Good job!

PR #867, opened 2022-11-16
sort students

MargaritaChorniak:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+): Student[] {
+  let newArray: Student[] = students.slice();
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (sortBy === 'grades') {
+    return newArray.sort((a: Student, b: Student): number => {
+      const valueA: number = a[sortBy].reduce(
+        (prev: number, curr: number) => prev + curr,
+      ) / a[sortBy].length;
```
TarasHoliuk, line 30: I'd create a separate function for this (since you use it 2 times)

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  if (sortBy === 'married') {
+    const arrA: Student[] = newArray
+      .filter((student: Student) => !!student.married);
+    const arrB: Student[] = newArray
+      .filter((student: Student) => !student.married);
+
+    return order === 'asc' ? arrB.concat(arrA) : arrA.concat(arrB);
+  }
+
+  if (order === 'asc') {
```
TarasHoliuk, line 49: You shouldn't have conditions by `order` and by `sortBy` on the same level
Use switch instead

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+        (prev: number, curr: number) => prev + curr,
+      ) / b[sortBy].length;
+
+      return order === 'asc' ? valueA - valueB : valueB - valueA;
+    });
+  }
+
+  if (sortBy === 'married') {
+    const arrA: Student[] = newArray
+      .filter((student: Student) => !!student.married);
```
TarasHoliuk, line 42: subtraction of the boolean value will return you the correct numbers

TarasHoliuk (changes requested):

Virass (approved):

PR #870, opened 2022-11-17
solution

oleksii-bieliaiev:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+            : b.age - a.age),
+      );
+
+    case SortType.Married:
+      return copy.sort(
+        (a: Student, b: Student) => (
+          order === 'asc'
+            ? Number(a.married) - Number(b.married)
+            : Number(b.married) - Number(a.married)),
+      );
```
loralevitska, line 54: you can combine it with helper function and pass need params, e.g. `sortByNumber`. In first case it would be `sortByNumber(a.age, b.age)` and second - `sortByNumber(Number(a.married), Number(b.married))`

loralevitska (changes requested):
great! let's improve

polosanya (approved):

PR #872, opened 2022-11-19
Solution ts_sort_students

anton-martyniuk1:


src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): string[] {
+  const studentsCopy = [...students];
+
+  if (sortBy === 'grades') {
```
TarasHoliuk, line 33: It's the perfect case to use `switch` construction

src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+    if (order === 'desc') {
```
TarasHoliuk, line 34: use ternary to return different values based on `order`

TarasHoliuk (changes requested):

Anastasiia-Svintsova (changes requested):

loralevitska (approved):
good job!

PR #882, opened 2022-11-26
solution

vladyslava-buzova:


tetiana-ivanchenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+    case SortType.AverageGrade:
+      sorting = (order === 'asc')
+        ? (st1: Student, st2: Student): number => getAverage(st1[sortBy])
+          - getAverage(st2[sortBy])
+        : (st1: Student, st2: Student): number => getAverage(st2[sortBy])
+          - getAverage(st1[sortBy]);
+      break;
+
+    default:
+      break;
```
tetiana-ivanchenko, line 64: default case should handling errors

tetiana-ivanchenko (changes requested):

PR #884, opened 2022-11-28
First Solution

DimaK123:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  const dublicate = [...students];
+
+  function average(array: number[]): number {
+    const sum = array.reduce((prev, current) => prev + current);
+
+    return sum / array.length;
+  }
+
+  if (order === 'asc') {
+    if (sortBy === SortType.Surname) {
```
tetiana-ivanchenko, line 34: use switch(sortBy)

tetiana-ivanchenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+  const dublicate = [...students];
+
+  function average(array: number[]): number {
+    const sum = array.reduce((prev, current) => prev + current);
+
+    return sum / array.length;
+  }
+
+  if (order === 'asc') {
+    switch (true) {
```
polosanya, line 33: You can just switch `sortBy`, as previous mentor said
Also, do order check inside cases

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+      case sortBy === SortType.Surname:
+        dublicate.sort((a, b) => a.surname.localeCompare(b.surname));
+        break;
+
+      case sortBy === SortType.Name:
+        dublicate.sort((a, b) => a.name.localeCompare(b.name));
```
polosanya, line 39: Think about combining the cases in which you are doing almost the same

polosanya (changes requested):
Please check the comments

src/sortStudents.ts
DimaK123: I did like you say but  i still think that i could leave check for asc/desc where they been and than use switch inside this check, so i don't need to do it all times in cases.

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+      } else {
+        dublicate.sort((a, b) => b.age - a.age);
+      }
+      break;
+
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        dublicate.sort((a, b) => average(a.grades) - average(b.grades));
+      } else {
+        dublicate.sort((a, b) => average(b.grades) - average(a.grades));
```
liliia-livandovskaya, line 57:
```suggestion
      return order === 'asc'
      ? dublicate.sort((a, b) => average(a.grades) - average(b.grades));
      : dublicate.sort((a, b) => average(b.grades) - average(a.grades));
```
That is not a mistake, but you also can use the [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

liliia-livandovskaya (approved):
Great job! 🔥

PR #892, opened 2022-12-06
solution

TetianaKoval:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  const allStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Age:
+      return allStudents.sort(
+        (a: Student, b: Student) => {
+          return order === 'asc'
+            ? +a[sortBy] - +b[sortBy]
```
liliia-livandovskaya, line 32:
```suggestion
            ? Number(a[sortBy]) - Number(b[sortBy])
```
That is not a mistake, but more preferable to use the Number()

PR #898, opened 2022-12-08
ts_sort_students

ysrckr:


src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+            (a: Student, b: Student) => {
+              const averageGradeA: number = a.grades
+                .reduce((accum: number, curr: number) => {
+                  return accum + curr;
+                }, 0) / a.grades.length;
+
+              const averageGradeB = b.grades
+                .reduce((accum: number, curr: number) => {
+                  return accum + curr;
+                }, 0) / b.grades.length;
```
mgrinko, line 85: This calculation is duplicated

src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+          });
+      }
+
+      return students
+        .map((student: Student) => student)
+        .sort((a: Student, b: Student) => {
+          return Number(b.married) - Number(a.married);
+        });
+    case SortType.AverageGrade:
+      if (order === 'asc') {
```
mgrinko, line 72: you don't need this `if` 3 times
you can just use `sign = 1 or -1` in the callback depending on the order

src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+        return students
+          .map((student: Student) => student)
+          .sort((a: Student, b: Student) => a.age - b.age);
+      }
+
+      return students
+        .map((student: Student) => student)
+        .sort((a: Student, b: Student) => b.age - a.age);
+    case SortType.Married:
```
mgrinko, line 57: I'd put a black line between cases

src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
...
+      if (order === 'asc') {
+        return students
+          .map((student: Student) => student)
+          .sort((a: Student, b: Student) => a.name.localeCompare(b.name));
+      }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      return students
+        .map((student: Student) => student)
```
mgrinko, line 35: why do you need this map?
the parameters already has a correct type

mgrinko (changes requested):

src/sortStudents.ts
ysrckr: I used map because it says "don't changed the original array"

src/sortStudents.ts
ysrckr: I couldn't do that because it doesn't pass the test so I used ternary.

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
 
+function averageGrade(grades: number[]): number {
```
naziamoff, line 2:
```suggestion
function getAverageGrade(grades: number[]): number {
```

naziamoff: function name should start with a verb

naziamoff (changes requested):
already looks good, but let's make a switch inside the sort callback - this will simplify your code a lot

naziamoff (changes requested):
What I meant - is that you have similar flow in a lot of cases. Now you're using switch for getting the order - but that is overkilling I guess.

You can combine the sorting of strings and numbers into single case

naziamoff (approved):
🔥

PR #900, opened 2022-12-09
solution

Fynkzia:


shyptia (approved):
Looks good!

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+      case SortType.Age:
+      case SortType.Married:
+        return Number(a[sortBy]) - Number(b[sortBy]);
+
+      case SortType.AverageGrade:
+        return getAverageNumber(a.grades) - getAverageNumber(b.grades);
+
+      default:
+        throw Error('This type doesn\'t exist');
+    }
```
ivan-kadykalo, line 51: Will be better to check the sort type first, and only then launch your sort depending on this sort type. Because now you are checking this type on every element during the whole iteration.

ivan-kadykalo (changes requested):
You are almost there 💪

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+};
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] {
+  let studentsCopy = [...students];
```
solaryasha, line 29: Avoid variable reassigning.

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+          ? getAverageNumber(firstPerson.grades)
+            - getAverageNumber(secondPerson.grades)
+          : getAverageNumber(secondPerson.grades)
+            - getAverageNumber(firstPerson.grades);
+      });
+
+      break;
+
+    default:
+      throw Error('This type doesn\'t exist');
```
solaryasha, line 64:
```suggestion
      throw Error(`This type doesn\'t exist : ${sortBy} `);
```

solaryasha (changes requested):

solaryasha (approved):
👍

PR #902, opened 2022-12-13
Solution

s3rhiienko:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  return grades.reduce((acum, current) => acum + current) / grades.length;
+}
+
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
+  return [...students].sort(
+    (student1: Student, student2: Student):number => {
+      if (sortBy === SortType.Name
+        || sortBy === SortType.Surname) {
+        return order === 'asc'
```
polosanya, line 30: image

polosanya (changes requested):
Good job! But let's try to remake this task using [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch), according to the checklist. It will improve your code

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+
+  switch (sortBy) {
+    case SortType.Age:
+      return allStudents.sort(
+        (a: Student, b: Student) => {
+          return order === 'asc'
+            ? +a[sortBy] - +b[sortBy]
+            : +b[sortBy] - +a[sortBy];
+        },
+      );
```
Anastasiia-Svintsova, line 34: Look, you repeat the same logic everywhere. Would be better to use a switch/case inside sort method. 
```
allStudents.sort(a: Student, b: Student) => {
  switch(sortBy)....

File: src/sortStudents.ts [hunk 3]
```diff
@@ -1,16 +1,68 @@
...
+          return order === 'asc'
+            ? averageValue(a[sortBy]) - averageValue(b[sortBy])
+            : averageValue(b[sortBy]) - averageValue(a[sortBy]);
+        },
+      );
 
+    case SortType.Married:
+      return allStudents
+        .filter((x) => x[sortBy] === true)
+        .concat(allStudents.filter((x) => x[sortBy] === false));
```
Anastasiia-Svintsova, line 63: I'm not sure that you need to use filter and concat students here. Just sort it like everywhere above

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #909, opened 2022-12-14
solution

lefteris-L:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+    if (sortBy === SortType.AverageGrade) {
+      const gradesA: Num[] = studentA.grades;
+      const gradesB: Num[] = studentB.grades;
+
+      const getSum = (x: Num, y: Num): Num => x + y;
+
+      const averageA: Num = gradesA.reduce(getSum) / gradesA.length;
+      const averageB: Num = gradesB.reduce(getSum) / gradesB.length;
+
+      return averageA - averageB;
```
ysrckr, line 41: You can make a function to calculate average grade so you wouldn't duplicate same things for each of them.

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
-  // write your function
-}
+  const sortCallback = (a: Student, b: Student): Num => {
+    let studentA: Student = a;
+    let studentB: Student = b;
+
+    if (order === 'desc') {
+      studentA = b;
+      studentB = a;
+    }
```
ysrckr, line 30: You can write with ternary so you wouldn't need to reassign.

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+      case SortType.Surname:
+        return studentA.surname.localeCompare(studentB.surname);
+
+      case SortType.Age:
+        return studentA.age - studentB.age;
+
+      case SortType.Married:
+        return !studentA.married && studentB.married
+          ? -1
+          : 1;
```
ysrckr, line 57: If you change boolean to number, you wouldn't need the ternary.

ysrckr (changes requested):

src/sortStudents.ts
lefteris-L: Great suggestions, thanks. Used all of them

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+      : b;
+
+    const studentB: Student = order === 'asc'
+      ? b
+      : a;
+
+    const isMarried = (student: Student): Num => (student.married
+      ? 1
+      : 0
+    );
```
ysrckr, line 35: You don't need this to convert `Boolean` to `Number`. 
You can write Number(studentA) to convert it to `Number`

ysrckr (approved):
LGTM 🤟🏻
Probably Nazar or Roman will ask for a few changes but I think this is good. 😄

src/sortStudents.ts
lefteris-L: Thanks, didn't know that

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+export type SortFunc = (a: Student[], b: SortType, c: SortOrder) => Student[];
```
meforoma, line 18: really nice 👏

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
-  // write your function
-}
+  const sortCallback = (a: Student, b: Student): Num => {
+    const studentA: Student = order === 'asc'
+      ? a
+      : b;
+
+    const studentB: Student = order === 'asc'
+      ? b
+      : a;
```
meforoma, line 30: you could make this to shorten the code
```suggestion
    const [studentA, studentB] = order === 'asc'
      ? [a, b]
      : [b, a];
```

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+    const getAverageGradeOf = (student: Student): Num => (
+      student.grades.reduce(getSum) / student.grades.length
+    );
+
+    switch (sortBy) {
+      case SortType.Name:
+        return studentA.name.localeCompare(studentB.name);
+
+      case SortType.Surname:
+        return studentA.surname.localeCompare(studentB.surname);
```
meforoma, line 42: these cases are essentially the same - let's join them
```suggestion
      case SortType.Name:
      case SortType.Surname:
        return studentA.surname.localeCompare(studentB.surname);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+
+      case SortType.Age:
+        return studentA.age - studentB.age;
+
+      case SortType.Married:
+        return Number(studentA.married) - Number(studentB.married);
```
meforoma, line 48: same here - join cases
```suggestion
      case SortType.Age:
      case SortType.Married:
        return Number(studentA.married) - Number(studentB.married);
```

meforoma (changes requested):
really un-trivial approach and that's awesome 👍 👍 👍 
let's just make the code bit shorter - take a look at few suggestions.
I'd actually approve already, but want to make sure you apply those optimizations 🙂

lefteris-L:
> really un-trivial approach and that's awesome 👍 👍 👍 let's just make the code bit shorter - take a look at few suggestions. I'd actually approve already, but want to make sure you apply those optimizations 🙂

Absolutely, will do

src/sortStudents.ts
lefteris-L: Awesome

PR #912, opened 2022-12-15
Solution

KrisMakarovska:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  Age,
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function Average(student: Student): number {
```
polosanya, line 20:
```suggestion
function getAverageGrades(student: Student): number {
```
Use camelCase. Also, please follow these recommendations on naming functions:
image

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+      studentA = b;
+      studentB = a;
+    }
+
+    switch (sortBy) {
+      case SortType.Name:
+        return studentA.name.localeCompare(studentB.name);
+
+      case SortType.Surname:
+        return studentA.surname.localeCompare(studentB.surname);
```
polosanya, line 45: You do the same in both cases. Think about combining them
```suggestion
      case SortType.Name:
      case SortType.Surname:
        return ....;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+      case SortType.Age:
+        return studentA.age - studentB.age;
+
+      case SortType.Married:
+        return (+studentA.married) - (+studentB.married);
```
polosanya, line 51: Same with age and married

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+      case SortType.AverageGrade:
```
polosanya, line 53: You can make it a default case

polosanya (changes requested):
Almost perfect! Let's improve your code a bit)

src/sortStudents.ts
KrisMakarovska: +

src/sortStudents.ts
KrisMakarovska: +

src/sortStudents.ts
KrisMakarovska: +

TarasHoliuk (approved):

PR #923, opened 2022-12-16
solution

Lun4trick:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return (order === 'asc')
+        ? studentsCopy.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : studentsCopy.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+
+    case (SortType.Age):
+    case (SortType.Married):
+      return (order === 'asc')
```
nurdanokcu, line 42:
```suggestion
    case SortType.Age:
    case SortType.Married:
      return order === 'asc'
```

I think parentheses aren't needed here and at the same other occurrences.

nurdanokcu (approved):
Neat, well done.

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+        : studentsCopy.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    case (SortType.AverageGrade):
+      return (order === 'asc')
+        ? studentsCopy.sort((a, b) => getAverageGrade(a) - getAverageGrade(b))
+        : studentsCopy.sort((a, b) => getAverageGrade(b) - getAverageGrade(a));
+    default:
+      throw new Error('Some error');
```
meforoma, line 51: better keep the error descriptive - in this case error would be thrown if `SortType` provided for the function is not supported - so Error should explain exactly this.

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+    case (SortType.Age):
+    case (SortType.Married):
+      return (order === 'asc')
+        ? studentsCopy.sort((a, b) => Number(a[sortBy]) - Number(b[sortBy]))
+        : studentsCopy.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    case (SortType.AverageGrade):
+      return (order === 'asc')
```
meforoma, line 47: agree with @nurdanokcu regarding the redundant parenthesis
```suggestion
    case SortType.AverageGrade:
      return order === 'asc'
```

meforoma (approved):
good 👍 
just consider few editorial comments

PR #925, opened 2022-12-16
solution

illiashcherbak:


src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+    case SortType.Name:
+    case SortType.Surname:
+      sortedStudentsList = sortedStudentsList
+        .sort((currStudent, nextStudent) => {
+          return order === 'asc'
+            ? currStudent[sortBy].localeCompare(nextStudent[sortBy])
+            : nextStudent[sortBy].localeCompare(currStudent[sortBy]);
+        });
+
+      break;
```
anastasiia-bilska, line 44:
```suggestion
  const sortedStudentsList = [...students];

  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return sortedStudentsList
        .sort((currStudent, nextStudent) => {
          return order === 'asc'
            ? currStudent[sortBy].localeCompare(nextStudent[sortBy])
            : nextStudent[sortBy].localeCompare(currStudent[sortBy]);
        });
```

anastasiia-bilska: sort is a mutable, you don't need to assign the result and you can use const for sortedStudentsList. You also can return sorted list, so you don't need to write 'break'. Same for next cases

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+
+    default:
+      break;
+  }
+
+  return sortedStudentsList;
 }
```
anastasiia-bilska, line 74:
```suggestion
    default:
       return sortedStudentsList;  
    }
}
```

anastasiia-bilska (changes requested):
Good! Please, check my comments

anastasiia-bilska (approved):
Well done!

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrades(gradesArray: number[]): number {
+  const sumOfGrades = gradesArray.reduce((acc, curr) => acc + curr, 0);
+  const avarageAge = sumOfGrades / gradesArray.length;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return avarageAge;
```
solaryasha, line 24:
```suggestion
  return sumOfGrades / gradesArray.length;;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudentsList = [...students];
```
solaryasha, line 32:
```suggestion
  const copiedStudents = [...students];
```

solaryasha (approved):

PR #929, opened 2022-12-17
Solution

Anastasia4sik:


NadiiaKoch (approved):
Good job!

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+            ? previous[sortBy].localeCompare(next[sortBy])
+            : next[sortBy].localeCompare(previous[sortBy]);
+
+        case (SortType.Age):
+          return order === 'asc'
+            ? previous[sortBy] - next[sortBy]
+            : next[sortBy] - previous[sortBy];
+
+        case (SortType.Married):
+          return order === 'asc'
```
ivan-kadykalo, line 45: In your solution, switch - case will be checked on every element throw the whole iteration. 
It will be better to check the sort type only once, before sorting.

ivan-kadykalo (approved):
Good job 🔥

PR #930, opened 2022-12-17
Solution

llllewvvaa:


RetaboKainrou (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+
+      case SortType.Age:
+      case SortType.Married:
+        return Number(a[sortBy]) - Number(b[sortBy]);
+
+      case SortType.AverageGrade:
+        return getAverageGrade(a.grades) - getAverageGrade(b.grades);
+
+      default:
+        throw new Error('Error: Unknown sort type');
```
ivan-kadykalo, line 51: Or we can return the same string without sorting

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+      : firstStudent;
+
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        return a[sortBy].localeCompare(b[sortBy]);
+
+      case SortType.Age:
+      case SortType.Married:
+        return Number(a[sortBy]) - Number(b[sortBy]);
```
ivan-kadykalo, line 45: Your algorithm launch switch case construction on each iteration to check the sortType. Would be better to check sort type firstly and later make sorting

ivan-kadykalo (changes requested):
You are almost there 💪

src/sortStudents.ts
llllewvvaa: **Fixed**

src/sortStudents.ts
llllewvvaa: **Fixed**

ivan-kadykalo (approved):
Good job 🔥

PR #936, opened 2022-12-18
solution

oleksandrnikora:


src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+const getAverageGrade = (grades: number[]): number => (
+  grades.reduce((sum, grade) => sum + grade, 0) / grades.length);
```
oleh-yablunovskyi, line 21:
```suggestion
const getAverageGrade = (grades: number[]): number => (
  grades.reduce((sum, grade) => sum + grade, 0) / grades.length
);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+          ? studentA[sortBy].localeCompare(studentB[sortBy])
+          : studentB[sortBy].localeCompare(studentA[sortBy]);
+
+      case SortType.Age:
+      case SortType.Married:
+        return order === 'asc'
+          ? Number(studentA[sortBy]) - Number(studentB[sortBy])
+          : Number(studentB[sortBy]) - Number(studentA[sortBy]);
 
+      default:
```
oleh-yablunovskyi, line 44:
```suggestion
      case SortType.AverageGrade:
```
Not sure it's a good idea to use default instead of specific case. At least in terms of code clarity. According to documentation, default is optional, so you can just ommit it.

oleh-yablunovskyi (approved):
Good job!

src/sortStudents.ts
solaryasha: Yeap, you are right. Let's use explicit case `case SortType.AverageGrade:`
By default, we don't sort or just throw an error

solaryasha (changes requested):

solaryasha (approved):
👍

PR #941, opened 2022-12-18
solution

kreisberg:


official-artem (approved):
Good job

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  return [...students].sort((student1: Student, student2: Student) => {
+    switch (sortBy) {
+      case SortType.Married:
+        return Number(student2[sortBy]) - Number(student1[sortBy]);
+
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'asc'
+          ? student1[sortBy].localeCompare(student2[sortBy])
+          : student2[sortBy].localeCompare(student1[sortBy]);
```
ivan-kadykalo, line 38: Switch case construction will be checked on each iteration.
Will be better to check the sort type first (before sort) and according to this type make a sort of your array.

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+        return order === 'asc'
+          ? getAverageGrades(student1)
+            - getAverageGrades(student2)
+          : getAverageGrades(student2)
+            - getAverageGrades(student1);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        return 0;
```
ivan-kadykalo, line 53: By default should be returned not sorter array

kreisberg: ✔️

src/sortStudents.ts
kreisberg: Fixes were added, if I get it correctly ✔️

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+        return order === 'asc'
+          ? getAverageGrades(student1)
+            - getAverageGrades(student2)
+          : getAverageGrades(student2)
+            - getAverageGrades(student1);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        throw new Error();
```
ivan-kadykalo, line 53: Describe what this error mean and why it throw

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+        return order === 'asc'
+          ? student1.age - student2.age
+          : student2.age - student1.age;
+
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAverageGrades(student1)
+            - getAverageGrades(student2)
+          : getAverageGrades(student2)
+            - getAverageGrades(student1);
```
ivan-kadykalo, line 50: You are checking your sort type every time, but it is defined only once. So will be better if you will define the sort type firstly and return sorted values next

```suggestion
  const sortedStudens = [...students];
  
    switch (sortBy) {
      case SortType.Married:
        return sortedStudens.sort((student1: Student, student2: Student) => (
          Number(student2.married) - Number(student1.married)
        );

      case SortType.Name:
      case SortType.Surname:
 ....

      case SortType.Age:
        return order === 'asc'
....

      case SortType.AverageGrade:
        return order === 'asc'
...
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/sortStudents.ts
kreisberg: ✔️

ivan-kadykalo (approved):
Good job 🔥

PR #942, opened 2022-12-18
sometimes I dream of writing the code of my life, but for now I just …

RedkoSerhii:
…sort students

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,69 @@
-
 export interface Student {
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAveregeGrade(studentGrades: number[]): number {
+  const sumOfGrades = studentGrades.reduce((sum: number, grade: number) => {
+    return sum + grade;
+  }, 0);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return sumOfGrades / studentGrades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  let copyStudents = [...students];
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyStudents = copyStudents.sort((student1, student2) => {
+        return order === 'asc'
+          ? student1[sortBy].localeCompare(student2[sortBy])
+          : student2[sortBy].localeCompare(student1[sortBy]);
+      });
+      break;
```
anastasiia-bilska, line 42:
```suggestion
  const copyStudents = [...students];

  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return copyStudents.sort((student1, student2) => {
        return order === 'asc'
          ? student1[sortBy].localeCompare(student2[sortBy])
          : student2[sortBy].localeCompare(student1[sortBy]);
      });
```

anastasiia-bilska: method sort is mutating, no need to use let and assign the result. also you can you can immediately return the value, so you don't need a break anymore. same with the next cases

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+            ? getAveregeGrade(student1.grades)
+              - getAveregeGrade(student2.grades)
+            : getAveregeGrade(student2.grades)
+              - getAveregeGrade(student1.grades);
+        });
+      break;
 
+    default:
+      break;
+  }
```
anastasiia-bilska, line 66:
```suggestion
    default:
			return copyStudents;
  }
```

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return copyStudents;
```
anastasiia-bilska, line 68: You don't need this if you use return in the switch

anastasiia-bilska (changes requested):
Good! Please check my cooments

src/sortStudents.ts
RedkoSerhii: Cool advice, thanks, fixed it.

src/sortStudents.ts
RedkoSerhii: fixed it

anastasiia-bilska (approved):
Good job!

solaryasha (approved):
👍

PR #947, opened 2022-12-19
Solution for TS task

RetaboKainrou:


src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+        case SortType.AverageGrade:
+          return order === 'asc'
+            ? getAverage(studentA[sortBy]) - getAverage(studentB[sortBy])
+            : getAverage(studentB[sortBy]) - getAverage(studentA[sortBy]);
+        default:
+          return 0;
+      }
```
llllewvvaa, line 52: There is a better to throw an error, than 0
```suggestion
        default:
          throw new Error('Unknown sort type');
      }
```

llllewvvaa (changes requested):

src/sortStudents.ts
RetaboKainrou: Fixed

PR #949, opened 2022-12-19
Add solution

VladimirEdelmann:


daryna-zinchenko (commented):
Great job! 🥇

daryna-zinchenko (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+      copyStudents = copyStudents.sort((first, second) => {
+        if (order === 'asc') {
+          return first[sortBy].localeCompare(second[sortBy]);
+        }
+
+        return second[sortBy].localeCompare(first[sortBy]);
+      });
+
+      break;
+
```
ivan-kadykalo, line 47: sort is a mutable method, so let's use const instead of let and return the result immediately
```suggestion
      return copyStudents.sort((first, second) => {
        if (order === 'asc') {
          return first[sortBy].localeCompare(second[sortBy]);
        }

        return second[sortBy].localeCompare(first[sortBy]);
      });

```

src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+          - findAverageGrade(first.grades);
+      });
+
+      break;
+
+    default:
+      break;
+  }
+
+  return copyStudents;
```
ivan-kadykalo, line 74:
```suggestion
    default:
      return copyStudents;
  }
```

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #950, opened 2022-12-19
Solution

CharAnagn:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Married = 'married',
+  AverageGrade = 'averageGrade'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+const getAvgGrade = (grades: number[]): number => {
+  return grades.reduce((a, b) => a + b, 0) / grades.length;
```
UmizDemud, line 22: This has a weak spot, if the grades array is empty there is a chance of having error on runtime or in devops/testing environments. I'd add something like

```js
  return grades.reduce((a, b) => a + b, 0) / ( grades.length || 1 );
```

UmizDemud (approved):
👍 

I'll approve, just consider my comment about that very small change

meforoma (approved):
agree with comment from @UmizDemud - it's totally valid
yet this implementation is approvable ✅

PR #956, opened 2022-12-19
added my solution

TheOneThatKnowsNothing:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(numbers: number[]): number {
```
TetKopot, line 20: You could use here grades  instead of number 
function getAverage(grades: number[]): number {

TetKopot (changes requested):
good, just a few changes need

src/sortStudents.ts
TheOneThatKnowsNothing: it's better with numbers, because we can use this function to get average value not only from grades

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+      case SortType.Name:
+        return order === 'asc'
+          ? firstStudent.name.localeCompare(secondStudent.name)
+          : secondStudent.name.localeCompare(firstStudent.name);
+
+      case SortType.Surname:
+        return order === 'asc'
+          ? firstStudent.surname.localeCompare(secondStudent.surname)
+          : secondStudent.surname.localeCompare(firstStudent.surname);
```
solaryasha, line 49: Use dynamic key `[sortBy]` and combine these cases

solaryasha (changes requested):
You should use the dynamic key `[sortBy]` in the switch instruction.

PR #957, opened 2022-12-19
Solution

alyonkaradchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  return [...students]
+    .sort((studentA: Student, studentB: Student) => {
+      switch (sortBy) {
+        case SortType.Name:
+        case SortType.Surname:
+          return studentA[sortBy].localeCompare(studentB[sortBy]);
```
viktoriia-kravchenko, line 34: You should also define sorting in descending order for names and surnames.

viktoriia-kravchenko (changes requested):
GJ! Only minor fixes should be made.

Ihor-Karpyn (approved):

PR #961, opened 2022-12-19
solution

grslckr:


src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
+        ? learner.sort((a, b) => +a[sortBy] - +b[sortBy])
+        : learner.sort((a, b) => +b[sortBy] - +a[sortBy]);
+
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? learner.sort((a, b) => getAverage(a.grades) - getAverage(b.grades))
+        : learner.sort((a, b) => getAverage(b.grades) - getAverage(a.grades));
+
+    default:
+      return [''];
```
UmizDemud, line 48: Why did you return an array with a single empty string 🧐

UmizDemud (approved):

src/sortStudents.ts
meforoma: yep, better throw a descriptive error e.g.
```suggestion
      throw new Error('SortType is not supported - ', sortBy);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
+    case SortType.Surname:
+      return order === 'asc'
+        ? learner.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : learner.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? learner.sort((a, b) => +a[sortBy] - +b[sortBy])
+        : learner.sort((a, b) => +b[sortBy] - +a[sortBy]);
```
meforoma, line 40: better don't use the `+` shorthand - but a Number casting instead
```suggestion
        ? learner.sort((a, b) => Number(a[sortBy]) - Number(b[sortBy]))
        : learner.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
```

meforoma (approved):
✅ 
but please checkup the comments

PR #967, opened 2022-12-20
Sort studens

maksym-kryvolap:


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
 
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: string): Object[] {
+  let copy: Array<Student> = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (order === 'asc') {
+    switch (sortBy) {
+      case 0:
```
TarasHoliuk, line 28:
```suggestion
      case SortType.Name:
```
use it in this way

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+        copy.sort((a, b) => a.name.localeCompare(b.name));
+        break;
+      case 1:
+        copy.sort((a, b) => a.surname.localeCompare(b.surname));
```
TarasHoliuk, line 32: you are doing the same in both cases - join it and use a computed property: `a[sortBy]`

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
```
TarasHoliuk, line 13:
```suggestion
  Name = 'name',
```
it will help you to join similar cases (you will be allow to use it as computed property)

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
+// import { objectExpression } from "@babel/types";
+// import { SignatureDeclarationBase } from "typescript";
```
TarasHoliuk, line 2: don't leave comments

TarasHoliuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  if (order === 'desc') {
+    const marriedPeople: Array<Student> = [];
+
+    switch (sortBy) {
+      case 'age':
+        copy.sort((a, b) => b.age - a.age);
+        break;
+      case 'married':
+
+        for (let i: number = 0; i < copy.length; i += 1) {
```
TarasHoliuk, line 54: sorry, missed it during the first review
you can transform a boolean to a number and use the common `sort` method here as well in this way:
`copy.sort((a, b) => +a.married - +b.married)`

TarasHoliuk (changes requested):
GJ! I've added one more suggestion (sorry, missed it during the first review)

nazarmatsevych (approved):
👍

PR #974, opened 2022-12-20
solution

mmahfuzi:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+        result = (a.grades.reduce((sum, grade) => sum + grade, 0)
+        / a.grades.length)
+        - (b.grades.reduce((sum, grade) => sum + grade, 0) / b.grades.length);
+        break;
+
+      default:
+        throw new Error('sort type not supported');
+    }
 
+    return order === 'asc' ? result : -result;
```
meforoma, line 56: not trivial but valid

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+        break;
+
+      case SortType.Married:
+        result = Number(a.married) - Number(b.married);
+        break;
+
+      case SortType.AverageGrade:
+        result = (a.grades.reduce((sum, grade) => sum + grade, 0)
+        / a.grades.length)
+        - (b.grades.reduce((sum, grade) => sum + grade, 0) / b.grades.length);
```
meforoma, line 49: would be better to create a helper function so you could shorten this part a lot
kind of
```suggestion
        result = getAverage(a.grades) - getAverage(b.grades);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  sortedStudents.sort((a, b) => {
+    let result = 0;
+
+    switch (sortBy) {
+      case SortType.Name:
+        result = a.name.localeCompare(b.name);
+        break;
+      case SortType.Surname:
+        result = a.surname.localeCompare(b.surname);
+        break;
```
meforoma, line 37: you can combine similar cases since the logic is pretty much the same
BUT for that you would have to refactor your enum so it allows you accessing the object properties - see the comment above
```suggestion
      case SortType.Name:
      case SortType.Surname:
        result = a[sortBy].localeCompare(b[sortBy]);
        break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
 }
 
 export enum SortType {
   // describe SortType enum
+  Name = 'n',
+  Surname = 's',
+  Age = 'a',
+  Married = 'm',
+  AverageGrade = 'aG',
```
meforoma, line 18:
```suggestion
  Name = 'name',//this will make it possible to access `student[SortType.Name]`
  Surname = 'surname',
  Age = 'age',
  Married = 'married',
  AverageGrade = 'aG',//here anything is fine, since there's no such prop on the Student object
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+      case SortType.Surname:
+        result = a.surname.localeCompare(b.surname);
+        break;
+      case SortType.Age:
+        result = a.age - b.age;
+        break;
+
+      case SortType.Married:
+        result = Number(a.married) - Number(b.married);
+        break;
```
meforoma, line 44: booleans can be converted to numbers 1 and 0
so you can combine these two cases too
```suggestion
      case SortType.Age:
      case SortType.Married:
        result = Number(a[sortBy]) - Number(b[sortBy]);
        break;
```

meforoma (changes requested):
there's a room for improvement @mmahfuzi

meforoma (approved):
✅

PR #978, opened 2022-12-21
sort students

Kavster92:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calcAverage(grades: number[]): number {
+  const sumOfGrades: number = grades.reduce((sum: number, current: number) => (
```
Grayfox147, line 21:
```suggestion
  return grades.reduce((sum, element) => sum + element, 0) / grades.length;
```
you could simplify the function, into a stright forward return with the value that you need.

Grayfox147 (commented):
Overall really good implementation of the code

meforoma (approved):
you could implement the @Grayfox147's suggestion
yet implementation is still approvable

PR #980, opened 2022-12-21
Develop

IoghiStudio:
Sorry for long solution.


meforoma (approved):
`switch..case` is preferable, but still approvable

PR #986, opened 2023-01-04
Sorting of the array according to the given params was implemented

mykyta-krainik:


src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+  return typeof value === 'string';
+}
+
+function isNumber(value: unknown): value is number {
+  return typeof value === 'number';
+}
+
+function isBoolean(value: unknown): value is boolean {
+  return typeof value === 'boolean';
+}
```
Anastasiia-Svintsova, line 29: You don't need these functions. Remove it please

src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+  ) {
+    return Number(propertyValueA) - Number(propertyValueB);
+  }
+
+  if (Array.isArray(propertyValueA) && Array.isArray(propertyValueB)) {
+    const averageGradeA = calculateAverageGrade(propertyValueA);
+    const averageGradeB = calculateAverageGrade(propertyValueB);
+
+    return averageGradeA - averageGradeB;
+  }
```
Anastasiia-Svintsova, line 59: Use switch/case here. You have sort types already, so use them instead of these type checking

src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+
+  // If all the above type checks fail
+  // the type of values is unknown
```
Anastasiia-Svintsova, line 62: Remove comments

Anastasiia-Svintsova (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  switch (sortBy) {
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? studentsCopy.sort((studentA, studentB) => {
+          return Number(studentA[sortBy]) - Number(studentB[sortBy]);
+        })
+        : studentsCopy.sort((studentA, studentB) => {
+          return Number(studentB[sortBy]) - Number(studentA[sortBy]);
+        });
```
Anastasiia-Svintsova, line 39: You don't need to call the sort function every time. Write it once and move the switch/case statement inside it, please

Anastasiia-Svintsova (changes requested):
You did a good job! Add some fixes, please

loralevitska (approved):
😎

PR #992, opened 2023-01-05
solution

euFarafonov:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
```
witflash, line 15: You can name all props with small letters

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copy: Student[] = [...students];
```
witflash, line 25: It's better to name like `sortedStudents`. `copy` doesn't say about what is it.

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
+      copy.sort((person1, person2) => {
+        return (order === 'asc')
```
witflash, line 31: You can do it with direct return for array function:
```
(person1, person2) => order === 'asc'
  ? // first condition
  : // second condition
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+        const avr1 = sum1 / person1.grades.length;
+
+        const sum2 = person2.grades.reduce((a, b) => (a + b));
+        const avr2 = sum2 / person2.grades.length;
+
+        return (order === 'asc')
+          ? (avr1 - avr2)
+          : (avr2 - avr1);
+      });
+      break;
```
witflash, line 58: This `break` is redundant I think

witflash (changes requested):

witflash (changes requested):
Good job!
Pay attention to the comments, please.

src/sortStudents.ts
euFarafonov: I can't name the properties with a small letter, because tests don't pass

witflash: Ok, no problem 👌🏼

witflash (approved):
👍🏻

PR #996, opened 2023-01-09
Solution v0.1

07Spirit:


src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+  order: keyof SortOrder,
+): Student[] {
+  const students = [...initialStudents];
 
-export function sortStudents(students, sortBy, order) {
   // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      students.sort((a, b) => {
+        if (order === 'asc') {
```
polosanya, line 36: Don't hardcore the value. Create an Enum and use it instead

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+
+        return 0;
+      });
+      break;
+
+    case SortType.Surname:
+      students.sort((a, b) => {
+        if (order === 'asc') {
+          return a.surname.localeCompare(b.surname);
+        }
```
polosanya, line 52: In both name and surname cases you have almost the same logic. Think about combining these cases

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+
+        return 0;
+      });
+      break;
+
+    case SortType.AverageGrade:
+      students.sort((a, b) => {
+        const aCalc: number = a.grades
+          .reduce((prev, sum) => prev + sum, 0) / a.grades.length;
+        const bCalc: number = b.grades
```
polosanya, line 90: Would be great to create a separate function for calculating average grades

polosanya (changes requested):
Almost done!

src/sortStudents.ts
```diff
@@ -1,16 +1,107 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export enum SortOrder {
+  ASC = 'asc',
+  DESC = 'desc',
+}
+
+function sortAvergeGrade(a: Student, b: Student, order: SortOrder): number {
+  const aCalc: number
```
loralevitska, line 26: Try to use more meaningful variable names. Now it's unclear `aCalc `

loralevitska (approved):

src/sortStudents.ts
07Spirit: Okay)

PR #1003, opened 2023-01-12
solution

vlad-trundenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function average(arr: number[]): number {
```
polosanya, line 20:
```suggestion
function getAverageGrades(arr: number[]): number {
```
Let's get used to name function according to these rules:
image

polosanya (approved):
Well done! Just please pay attention to my comment)

PR #1007, opened 2023-01-15
Solution

andrieieva-yuliia:


OlhaYevstifieieva (approved):
Looks nice 👌

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+        },
+      );
+
+    case SortType.AverageGrade:
+      return copiedStudents.sort(
+        (a: Student, b: Student) => {
+          function averageValue(arr: number[]): number {
+            return arr.reduce((prevNum, nextNum) => prevNum + nextNum)
+            / arr.length;
+          }
```
Anna-Tkachenko, line 53: will be better to move additional function outside of the `sortStudents` function
also, function name should contain an action, I don't see any action name

Anna-Tkachenko: also don't use data type name for variable names

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copiedStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Age:
+      return copiedStudents.sort(
```
Anna-Tkachenko, line 29: let's just sort needed array for each case, and will return the result array at the end of the function one time
in this way your code will be clearer

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+          return order === 'asc'
+            ? averageValue(a[sortBy]) - averageValue(b[sortBy])
+            : averageValue(b[sortBy]) - averageValue(a[sortBy]);
+        },
+      );
+
+    case SortType.Married:
+      return copiedStudents
+        .filter((x) => x[sortBy] === true)
+        .concat(copiedStudents.filter((x) => x[sortBy] === false));
```
Anna-Tkachenko, line 64: check how we can sort array for boolean field via sort method (through 1 iteration instead of 3)

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1012, opened 2023-01-16
add solution

Anastasiia-Khudych:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function getAverageGrade(grades: number[]): number {
+  return grades.reduce((accum, currentValue) => {
+    return accum + currentValue;
+  }, 0) / grades.length;
+}
```
Avichez, line 23: You can write it into one line but if here is more than 80 symbols, then, better to move the method to the next line.
```suggestion
function getAverageGrade(grades: number[]): number {
  return grades
     .reduce((accum, currentValue) => accum + currentValue, 0) / grades.length;
}
```

Avichez (approved):
Great solution 🔥 
Short and understandble 🙂

src/sortStudents.ts
ArturErmolenko: 👍

ArturErmolenko (approved):
GJ 👍

PR #1015, opened 2023-01-17
Solution

oleh-holovnykh:


src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
-export function sortStudents(students, sortBy, order) {
-  // write your function
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  return [...students].sort((a, b) => {
+    switch (sortBy) {
+      case ('name'):
```
svitjojo, line 32: for cases, you can use your enum. Otherwise, enum is useless, you don't use it.
```suggestion
      case SortType.Name:
```

svitjojo (changes requested):
Good job 👍 
Use your enum on maximum)

src/sortStudents.ts
oleh-holovnykh: Thnx! Changed. Enum on maximum now 😎

Anna-Tkachenko (approved):
Good job!

PR #1016, opened 2023-01-17
Sort students TS - completed

GarryRocksman:
Have a nice day 🔥

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,82 @@
 
 export interface Student {
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(grades: number[]): number {
+  return grades.reduce((prev: number, curr: number) => (
+    prev + curr
+  )) / grades.length;
+}
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+function sortString(a: string, b: string, order: string): number {
+  return order === 'asc'
+    ? a.localeCompare(b)
+    : b.localeCompare(a);
+}
+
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+      studentsCopy.sort((a: Student, b: Student) => (
+        sortString(a.name, b.name, order)
+      ));
+      break;
+
+    case SortType.Surname:
+      studentsCopy.sort((a: Student, b: Student) => (
+        sortString(a.surname, b.surname, order)
+      ));
+      break;
```
Anastasiia-Khudych, line 57: You can combine these if you set the values for enum properties like this `Name = 'name'`. And then use `studentsCopy.sort((a: Student, b: Student) => (
        sortString(a[sortBy], b[sortBy], order)
      ));`

Anastasiia-Khudych (changes requested):
Good job!

src/sortStudents.ts
GarryRocksman: thanks, fixed 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grade',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(grades: number[]): number {
```
Anna-Tkachenko, line 21: function name should contain some action name

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+  return grades.reduce((prev: number, curr: number) => (
+    prev + curr
+  )) / grades.length;
+}
+
+function sortString(a: string, b: string, order: string): number {
```
Anna-Tkachenko, line 27: use enum as the type for order argument

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+  return order === 'asc'
+    ? a.localeCompare(b)
+    : b.localeCompare(a);
+}
+
+function sortNumber(a: number, b: number, order: string): number {
```
Anna-Tkachenko, line 33: same here about `order`

Anna-Tkachenko (approved):

PR #1019, opened 2023-01-17
sortStudents function was added

svitjojo:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageMarks(marks: number[]): number {
+  return marks
+    .reduce((total: number, current: number): number => total + current, 0)
```
ArturErmolenko, line 23: You don't need this types, array already typified above ⬆️ 
```suggestion
    .reduce((acc ,current): number => acc + current, 0)
```

ArturErmolenko (approved):
Well done 👍

PR #1024, opened 2023-01-18
Solution

metalikaqq:
not so ez


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAver = (marks:number[]): number => {
```
Avichez, line 20: Try to give a full name to the function.

Avichez (approved):
Great solution 🙂 👍

src/sortStudents.ts
ArturErmolenko, line 20: bad naming
```suggestion
const getAvarageGrades = (greads:number[]): number => {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  const averageMark:number = marks
+    .reduce((prev: number, curr: number): number => prev + curr, 0)
+    / marks.length;
```
ArturErmolenko, line 23: You can return immediately
You don't need typify callback , you are working with  array that is already typed
```suggestion
  return grades
    .reduce((acc, curr):  => acc + curr, 0)
    / marks.length;
```

ArturErmolenko (changes requested):
GJ 👍 Pay attention to the variable names
Left comments

Denchepornyuk (approved):

PR #1028, opened 2023-01-18
solution

imariash86:


src/sortStudents.ts
```diff
@@ -1,16 +1,118 @@
...
+      break;
+
+    case 'grades':
+      res = [...students].sort((a: Student, b: Student): number => {
+        return order === 'desc'
+          ? b[sortBy].reduce((acc, cur) => acc + cur) / b[sortBy].length
+            - a[sortBy].reduce((acc, cur) => acc + cur) / a[sortBy].length
+          : a[sortBy].reduce((acc, cur) => acc + cur) / a[sortBy].length
+            - b[sortBy].reduce((acc, cur) => acc + cur) / b[sortBy].length;
+      });
```
NVyrsta, line 64: It is better to create a separate function for repeated calculations

NVyrsta (changes requested):
Nice work, but try to create a function to calculate the average values to avoid code duplication

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
   // write your function
+  let res: Student[] = [];
+
+  switch (sortBy) {
+    case 'age':
```
ArturErmolenko, line 32: Why you don't use `enum`
```suggestion
    case SortType.Age:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+    case 'married':
+      res = [...students].sort((a: Student, b: Student): number => {
```
ArturErmolenko, line 34: lets follow DRY , you repeat it 3 times , les't return immediately array  and sort it just one time

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+          return arr.reduce((acc, cur) => acc + cur, 0) / arr.length;
+        }
+
+        return order === 'desc'
+          ? calc(b[sortBy]) - calc(a[sortBy])
+          : calc(a[sortBy]) - calc(b[sortBy]);
+      });
+      break;
+    default:
+      throw new Error('There is an error!');
```
ArturErmolenko, line 62: what kind of error , error should give us some info :)

ArturErmolenko (changes requested):

Denchepornyuk (approved):

PR #1029, opened 2023-01-18
solution

Anton-Zhytonbaiev:


OlhaYevstifieieva (approved):
GJ 🫂

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAver = (marks:number[]): number => {
```
ArturErmolenko, line 20: use better naming
```suggestion
const getAvaregeGrades = (marks:number[]): number => {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  const averageMark:number = marks
+    .reduce((prev: number, curr: number): number => prev + curr, 0)
+    / marks.length;
```
ArturErmolenko, line 23: 1. You can `return` immediately
2. `Array` is already typed above so you don't need type callback arguments

```suggestion
  return marks
    .reduce((acc, curr): number => acc + curr, 0)
    / marks.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return averageMark;
```
ArturErmolenko, line 25:
```suggestion
```

ArturErmolenko (changes requested):
GJ 👍 
Left comments ⭐

ArturErmolenko (approved):
Well Done 👍

PR #1030, opened 2023-01-18
v-01

AndreyBerezyanskiy:


tania-boiar (approved):
Good job! 😊

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(students: Student[], sortBy: SortType,
+  order: SortOrder):Student[] {
```
Anna-Tkachenko, line 21: move each argument to the new line

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  type CopyObjectCallack = (obj: Student) => Student;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const copyObjectCallack: CopyObjectCallack = (obj) => ({ ...obj });
```
Anna-Tkachenko, line 24:
```suggestion
  const copyObjectCallback: CopyObjectCallack = (obj) => ({ ...obj });
```

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  const copyStudents: Student[] = students.map(copyObjectCallack);
```
Anna-Tkachenko, line 25: it's redundant here to make a copy of each object in the new array
will be enough just create copy of array, because you won't change objects in array, you just will change their position

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+          ? Number(a.married) - Number(b.married)
+          : Number(b.married) - Number(a.married);
+      });
+      break;
+
+    case SortType.AverageGrade:
+      copyStudents.sort((a: Student, b: Student) => {
+        const aGradeAverage: number = a.grades
+          .reduce((sum:number, element: number) => sum + element, 0)
+          / a.grades.length;
```
Anna-Tkachenko, line 57: don't repeat yourself: create separate function that will calculate average grade and return you needed number, that you'll be able to use in sort function

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+          .reduce((sum:number, element: number) => sum + element, 0)
+          / b.grades.length;
+
+        return order === 'asc'
+          ? aGradeAverage - bGradeAverage
+          : bGradeAverage - aGradeAverage;
+      });
+      break;
+
+    default:
```
Anna-Tkachenko, line 69: will be better to throw an error with understandable message in default case

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1032, opened 2023-01-18
Solution

BrandonSunrise:


anna-inozemets (approved):
Pretty good solution🐇

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+const getAverage = (marks:number[]): number => {
+  const averageMark:number = marks
+    .reduce((prevNum: number, nextNum: number): number => prevNum + nextNum, 0)
+    / marks.length;
+
+  return averageMark;
+};
```
ArturErmolenko, line 25: You can return immideatly
```suggestion
const getAverageGrades = (marks:number[]): number => {
  return marks
    .reduce((prevNum, nextNum): number => prevNum + nextNum, 0)
    / marks.length;

  return averageMark;
};
```

ArturErmolenko (approved):
Well done 👍

PR #1033, opened 2023-01-18
solution

victorionov:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+          ? a.name.localeCompare(b.name)
+          : b.name.localeCompare(a.name);
+      });
+
+    case SortType.Surname:
+      return myStudents.sort((a: Student, b: Student) => {
+        return isOrderAscending
+          ? a.surname.localeCompare(b.surname)
+          : b.surname.localeCompare(a.surname);
+      });
```
Anastasiia-Khudych, line 48: You can combine these two if change `a.surname` and `b.surname` with `a[sortBy]` and `b[sortBy]`

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const myStudents = students.map((student: Student) => ({ ...student }));
```
Anastasiia-Khudych, line 25: You can use destructuring here because objects will not be changed, only sorted.

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      return myStudents.sort((a: Student, b: Student) => {
+        const avgA: number = a.grades
+          .reduce((acc: number, x: number) => acc + x, 0) / a.grades.length;
+        const avgB: number = b.grades
+          .reduce((acc: number, x: number) => acc + x, 0) / b.grades.length;
+
+        return isOrderAscending
+          ? avgA - avgB
+          : avgB - avgA;
+      });
```
Anastasiia-Khudych, line 67: I suppose it's not the default case.

Anastasiia-Khudych (changes requested):
Well done, but check the comments

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const myStudents = [...students];
```
Anna-Tkachenko, line 25: why the student is yours? 🙂

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+      return myStudents.sort((a: Student, b: Student) => {
+        return isOrderAscending
+          ? Number(a.married) - Number(b.married)
+          : Number(b.married) - Number(a.married);
+      });
+
+    default:
+      return myStudents.sort((a: Student, b: Student) => {
+        const avgA: number = a.grades
+          .reduce((acc: number, x: number) => acc + x, 0) / a.grades.length;
```
Anna-Tkachenko, line 54: don't repeat yourself: let's create separate function that will calculate the average grade value, that you'll be able to use in sort function

src/sortStudents.ts
Anna-Tkachenko: agree, fix @Anastasiia-Khudych , it's not the default case, it's case for some field

Anna-Tkachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+function getAverageGrade(grades: number[]): number {
+  return grades.reduce((a, b) => a + b) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const StudentsCopy = [...students];
```
Anna-Tkachenko, line 29: don't start simple variables with capital letter
will be better to use it only for some constants, that you'll use through the project
```suggestion
  const studentsCopy = [...students];
```

Anna-Tkachenko (approved):
Good job!

PR #1037, opened 2023-01-18
Solution

MFrolova-art:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+          : b[sortBy] - a[sortBy];
+      });
+      break;
+
+    case SortType.Married:
+      copyStudents.sort((a, b) => {
+        return order === 'asc'
+          ? Number(a[sortBy]) - Number(b[sortBy])
+          : Number(b[sortBy]) - Number(a[sortBy]);
+      });
```
svitjojo, line 51: You can combine these two cases
```suggestion
    case SortType.Age:
    case SortType.Married:
      copyStudents.sort((a, b) => {
        return order === 'asc'
          ? Number(a[sortBy]) - Number(b[sortBy])
          : Number(b[sortBy]) - Number(a[sortBy]);
      });
```

svitjojo (changes requested):
Good job 👍 
I think also you can remove 'break' because 'return' do the same.

PR #1040, opened 2023-01-18
Solution

chermak-oleh:


svitjojo (approved):
Great job 👍

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAverage = (grades: number[]): number => {
```
Anna-Tkachenko, line 20: will be better to add some noun in function naming, for example `getAverageValue`

Anna-Tkachenko: or `getAverageGrade` if you use `grades` as the name for the argument

Anna-Tkachenko (approved):

PR #1043, opened 2023-01-18
Ts_sort_students

SerhiiVovchenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
+type Num = number;
+type Str = string;
```
qoa11a, line 2: there is no need to create types for the primitives

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+      });
+      break;
+
+    case SortType.Surname:
+      copyStudent.sort((a, b) => {
+        return (order === 'asc')
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+      });
+      break;
```
qoa11a, line 48: you have exactly same logic for these 2 cases. You should not repeat yourself

qoa11a (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function averageGrade(grades: number[]) : number {
```
polosanya, line 19:
```suggestion
function getAverageGrade(grades: number[]) : number {
```
Get used to name functions according to these rules:
image

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  return grades.reduce((a: number, b: number) => a + b, 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyStudent: Student[] = [...students];
```
```suggestion
  const copyOfStudents: Student[] = [...students];
```
It's a copy of all students, not just one

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      copyStudent.sort((a, b) => {
+        return (order === 'asc')
+          ? Number(a[sortBy]) - Number(b[sortBy])
+          : Number(b[sortBy]) - Number(a[sortBy]);
+      });
+      break;
+
+    case SortType.AverageGrade:
+
+      copyStudent.sort((a, b) => {
```
polosanya, line 58:
```suggestion
    case SortType.AverageGrade:
      copyStudent.sort((a, b) => {
```

polosanya (changes requested):
Almost perfect!

SerhiiVovchenko:
Thank you very much for your comments!)

polosanya (approved):
Well done!

PR #1045, opened 2023-01-18
task sort students was implemented

OlyaZaiets:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+
+      break;
+
+    case SortType.Married:
+      copiedStudents.sort((currentStudent, nextStudent) => {
+        return order === 'asc'
+          ? +currentStudent[sortBy] - (+nextStudent[sortBy])
+          : +nextStudent[sortBy] - (+currentStudent[sortBy]);
+      });
```
nzrii-kysel, line 53: 😻you😻can😻combine😻these😻cases😻

nzrii-kysel (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function findAverageGrade(grades: number[]): number {
+  return grades.reduce((sum, mark) => sum + mark, 0) / grades.length;
+}
+
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
```
ArturErmolenko, line 25:
```suggestion
export function sortStudents(
  students: Student[],
  sortBy: SortType, 
  order: SortOrder
  ): Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copiedStudents.sort((currentStudent, nextStudent) => {
+        return order === 'asc'
+          ? currentStudent[sortBy].localeCompare(nextStudent[sortBy])
+          : nextStudent[sortBy].localeCompare(currentStudent[sortBy]);
+      });
+
+      break;
```
ArturErmolenko, line 37: Let's try follow DRY , `copiedStudents.sort` repeat three times , can we make return immediately like:
```suggestion
return copiedStudents.sort((a, b) => {
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
        return order === 'asc'
          ? a[sortBy].localeCompare(b[sortBy])
          : b[sortBy].localeCompare(a[sortBy]);
```

1. also you don't need brake after `return`
2. and with other cases the same, it make your code more readable and you use sort just once :)

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+          ? findAverageGrade(currentStudent[sortBy])
+            - findAverageGrade(nextStudent[sortBy])
+          : findAverageGrade(nextStudent[sortBy])
+            - findAverageGrade(currentStudent[sortBy]);
+      });
+
+      break;
+
+    default:
+      throw new Error('Oops... you tried to sort by non-existent type');
```
ArturErmolenko, line 61: what if I pass incorrect array? better use `args` or smith else

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return copiedStudents;
```
ArturErmolenko, line 64:
```suggestion
```

ArturErmolenko (changes requested):
Almost done 👍

PR #1048, opened 2023-01-18
Solution

vanst1:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+        .sort((a, b) => (b[sortBy].reduce((x, y) => x + y, 0)
+        / b[sortBy].length)
+        - (a[sortBy].reduce((x, y) => x + y, 0) / a[sortBy].length));
+
+    default:
+      if (order === 'asc') {
+        return [...students].sort((a, b) => +a[sortBy] - +b[sortBy]);
+      }
+
+      return [...students].sort((a, b) => +b[sortBy] - +a[sortBy]);
```
GarryRocksman, line 50: i don't think that default is a right please for one more sorting type, it's better to use it like an error catcher if some unusual 'order' will be passed

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      if (order === 'asc') {
+        return [...students].sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+      }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      return [...students].sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```
GarryRocksman, line 30: you can put sort  in a separate function and use it here, you can combine sort methods such as string sort and numbers sort to prevent code functional duplication

GarryRocksman (approved):
GJ 🔥 But needs a little fixes

src/sortStudents.ts
ArturErmolenko, line 50: here should be Error

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+        return [...students]
+          .sort((a, b) => (a[sortBy].reduce((x, y) => x + y, 0)
+          / a[sortBy].length)
+          - (b[sortBy].reduce((x, y) => x + y, 0) / b[sortBy].length));
+      }
+
+      return [...students]
+        .sort((a, b) => (b[sortBy].reduce((x, y) => x + y, 0)
+        / b[sortBy].length)
+        - (a[sortBy].reduce((x, y) => x + y, 0) / a[sortBy].length));
```
ArturErmolenko, line 43: let us use ternary here

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+        return [...students].sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+      }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      return [...students].sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+
+    case SortType.AverageGrade:
+      if (order === 'asc') {
+        return [...students]
```
ArturErmolenko, line 34: you already sort array above in first return , follow DRY

ArturErmolenko (changes requested):
GJ 👍 
Left comments ⭐

alexandra-protyanova (approved):
Good job 👏

PR #1049, opened 2023-01-18
sollution

serbinm:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+        if (order === 'asc') {
+          return a[sortBy].localeCompare(b[sortBy]);
+        }
+
+        return b[sortBy].localeCompare(a[sortBy]);
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAvgGrade(a[sortBy]) - getAvgGrade(b[sortBy])
+          : getAvgGrade(b[sortBy]) - getAvgGrade(a[sortBy]);
+      default: return [...students];
```
vanst1, line 55: here should be Error
` default:
      throw new Error('Error message');`

stadnytskyi-ivan (changes requested):
Good job:)
Some fix needed

PR #1054, opened 2023-01-19
solution

vladyslav78292:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+    case SortType.AverageGrade:
+      return studentsCopyArray.sort((a, b) => {
+        return order === 'asc'
+          ? getAverageMark(a[sortBy]) - getAverageMark(b[sortBy])
+          : getAverageMark(b[sortBy]) - getAverageMark(a[sortBy]);
+      });
+
+    default:
+      throw new Error('Error');
+  }
```
PolinaSukhorukova, line 69: You can merge sorting of fields with string and number(with boolean) types.

PolinaSukhorukova (approved):
You can merge

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'desk' | 'asc';
```
Anna-Tkachenko, line 18: check correct value for sort order type

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
 
+function getAverageMark(marksArray: number[]): number {
```
Anna-Tkachenko, line 20: don't use data type name in variable name

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  return marksArray.reduce((sum, el) => sum + el, 0) / marksArray.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] | string {
+  const studentsCopyArray: Student[] = [...students];
```
Anna-Tkachenko, line 29: don't use data type name in variable name

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+      return order === 'asc'
+        ? studentsCopyArray.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : studentsCopyArray.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+
+    case SortType.Surname:
+      studentsCopyArray.sort((a, b) => {
+        return order === 'asc'
```
Anna-Tkachenko, line 39: check point 3 in checklist

Anna-Tkachenko (changes requested):

src/sortStudents.ts
vladyslav78292: There is no variable here, I removed data typein `studentsCopyArray` variable

Anna-Tkachenko: I'm about `marksArray`, you created separate variable as an function argument, don't use data type name here

Anna-Tkachenko (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+
+    case SortType.AverageGrade:
+      return studentsCopyArray.sort((a, b) => {
+        return order === 'asc'
+          ? getAverageMark(a[sortBy]) - getAverageMark(b[sortBy])
+          : getAverageMark(b[sortBy]) - getAverageMark(a[sortBy]);
+      });
+
+    default:
+      throw new Error('Error');
```
Anna-Tkachenko, line 64: Specify understandable error text instead of just `error` string

PR #1063, opened 2023-01-22
Students sorting function has been created

Oleksandr-Antonenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,49 @@
...
+          : b[sortBy].localeCompare(a[sortBy]);
 
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAverageGrade(a[sortBy]) - getAverageGrade(b[sortBy])
+          : getAverageGrade(b[sortBy]) - getAverageGrade(a[sortBy]);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
```
tetiana-ivanchenko, line 43: default value should handling error instead of do smth for valid cases
create cases:
    case SortType.Age:
    case SortType.Married:

tetiana-ivanchenko (changes requested):

vlad-demchuk (approved):

PR #1073, opened 2023-01-30
sorted students

Mykola-Busarhin:


loralevitska (changes requested):
Good job!
Let's improve
1. Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others
2. DRY - Combine repetitive logic into helper functions (like all numbers/strings, or asc/desc)

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+  let result: object[];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      result = order === 'asc'
+        ? copyOfPeople.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : copyOfPeople.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+      break;
+    case SortType.Age:
```
tetiana-ivanchenko, line 38: combine with maried case

tetiana-ivanchenko (changes requested):

SanyaBratashchuk (approved):

PR #1075, opened 2023-02-01
Finish task

U-Ra-All:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+        return order === 'asc'
+          ? studentA.name.localeCompare(studentB.name)
+          : studentB.name.localeCompare(studentA.name);
+
+      case SortType.Surname:
+        return order === 'asc'
+          ? studentA.surname.localeCompare(studentB.surname)
+          : studentB.surname.localeCompare(studentA.surname);
```
OlEzhA0, line 34: you can modify your enum like 
```
export enum SortType {
  Name = 'name',
  Surname = 'surname',
  Age = 'age',
  Married = 'married',
  AverageGrade = 'grades',
}
```
and use your switch like
```
      case SortType.Name:
      case SortType.Surname:
        return order === 'asc'
          ? studentA[sortBy].localeCompare(studentB[sortBy])
          : studentB[sortBy].localeCompare(studentA[sortBy]);
```

OlEzhA0 (changes requested):

src/sortStudents.ts
U-Ra-All: Ok, done

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+        };
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+        return order === 'asc'
+          ? studentA[sortBy].reduce(callback, 0)
+          - studentB[sortBy].reduce(callback, 0)
+          : studentB[sortBy].reduce(callback, 0)
+          - studentA[sortBy].reduce(callback, 0);
+      }
```
etojeDenys, line 56: we can create two vars for better readability

```
 const averageGradeA: number = studentA.grades.reduce(
      (acc: number, grade: number) => acc + grade,
    ) / studentA.grades.length;
    const averageGradeB: number = studentB.grades.reduce(
      (acc: number, grade: number) => acc + grade,
    ) / studentB.grades.length;
```

etojeDenys: and your `return` will look much more simple

etojeDenys (changes requested):

src/sortStudents.ts
U-Ra-All: Ok, done

IvanFesenko (approved):

PR #1076, opened 2023-02-01
sort students

katerinalex:
sort students

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,127 @@
 
 export interface Student {
-  // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  let resultSt: Student[] = [];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Age:
+      resultSt = [...students].sort((a: Student, b: Student) => {
+        const numA: number = a.age;
+        const numB: number = b.age;
+
+        if (order === 'asc') {
+          return numA - numB;
+        }
+
+        return numB - numA;
+      });
+      break;
+
+    case SortType.Name:
+      resultSt = [...students].sort((a: Student, b: Student) => {
+        let stringA: string;
+        let stringB: string;
+
+        if (order === 'asc') {
+          stringA = a.name;
+          stringB = b.name;
+        } else {
+          stringA = b.name;
+          stringB = a.name;
+        }
+
+        if (stringA > stringB) {
+          return 1;
+        }
+
+        if (stringA === stringB) {
+          return 0;
+        }
+
+        return -1;
+      });
+      break;
+
+    case SortType.Surname:
+      resultSt = [...students].sort((a: Student, b: Student) => {
+        let stringA: string;
+        let stringB: string;
+
+        if (order === 'asc') {
+          stringA = a.surname;
+          stringB = b.surname;
+        } else {
+          stringA = b.surname;
+          stringB = a.surname;
+        }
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  let resultSt: Student[] = [];
```
polosanya, line 25: You can make it a copy of students, so you don't need to reassign it later.
Also, try to use more descriptive names

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Age:
+      resultSt = [...students].sort((a: Student, b: Student) => {
```
polosanya, line 29: image

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
+        const numA: number = a.age;
+        const numB: number = b.age;
+
+        if (order === 'asc') {
+          return numA - numB;
+        }
+
+        return numB - numA;
```
polosanya, line 37: You can use ternary operator here.

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
+
+        if (stringA > stringB) {
+          return 1;
+        }
+
+        if (stringA === stringB) {
+          return 0;
+        }
+
+        return -1;
```
polosanya, line 62: You can't compare strings like that. Use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) method to compare them correctly

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
```
polosanya, line 87: Same here. Also, you do almost the same in both cases. Think about combining them

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
+      });
+      break;
+
+    case SortType.Married:
+      resultSt = [...students].sort((a: Student, b: Student) => {
+        const numA: number = +a.married;
+        const numB: number = +b.married;
+
+        if (order === 'asc') {
+          return numA - numB;
```
polosanya, line 97: Same with Married and Age

src/sortStudents.ts
```diff
@@ -1,16 +1,127 @@
...
+
+    case SortType.AverageGrade:
+      resultSt = [...students].sort((a: Student, b: Student) => {
+        const numA: number = (a.grades.reduce(
+          (accumulator, currentValue) => accumulator + currentValue, 0,
+        )) / a.grades.length;
+
+        const numB: number = (b.grades.reduce(
+          (accumulator, currentValue) => accumulator + currentValue, 0,
+        )) / b.grades.length;
```
polosanya, line 112: Would be great to create a separate function for calculating average grades

polosanya (changes requested):
Almost done! Let's improve your code a bit)

src/sortStudents.ts
```diff
@@ -1,16 +1,106 @@
...
+          stringA = a.surname;
+          stringB = b.surname;
+        } else {
+          stringA = b.surname;
+          stringB = a.surname;
+        }
+
+        return stringA.localeCompare(stringB);
+      });
+      break;
```
Anastasiia-Svintsova, line 81: You do the same things here. Combine these 2 cases, please

src/sortStudents.ts
```diff
@@ -1,16 +1,106 @@
...
+
+    case SortType.Married:
+      resultSt.sort((a: Student, b: Student) => {
+        const numA: number = +a.married;
+        const numB: number = +b.married;
+
+        return (order === 'asc' ? numA - numB : numB - numA);
+      });
+      break;
```
Anastasiia-Svintsova, line 90: Can you combine this case with the age?

Anastasiia-Svintsova (changes requested):
- Also you can write the sort method only once and write the switch/case inside it 
```
[...students].sort((a: Student, b: Student) => {
  switch (sortBy) {
    case...
  }
})

qoa11a (approved):

PR #1088, opened 2023-02-11
commit

skomskiy:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+function getAverage(arr: number[]): number {
+  return arr.reduce((acc: number, cur: number) => acc + cur) / arr.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copy = [...students];
```
Volodymyr-Melnyk-Fron10der, line 30: copy of what? which copy?)
```suggestion
  const copiedStudents = [...students];
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? copy.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : copy.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+
```
Volodymyr-Melnyk-Fron10der, line 38: **(Optional)** No problem with your solution here, but just FYI there is more advanced way to deal with case like that:

This solution is more flexible in case of growth of functionality 
```suggestion

const getSortCallbacks = (sortBy: SortType) => {
  asc: (a, b) => a[sortBy].localeCompare(b[sortBy]),
  desc: (a, b) => b[sortBy].localeCompare(a[sortBy])
}

.....

  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return copy.sort(getSortCallbacks(sortBy)[order]);

```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(arr: number[]): number {
```
Volodymyr-Melnyk-Fron10der, line 21: Actually this type is useless, because TS will set this type by default based on function return type
```suggestion
function getAverage(arr: number[]) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, almost there👍 A few comment to improve:

slavanosatskyi (approved):
well done!

etojeDenys (approved):

PR #1101, opened 2023-02-17
initial commit

ihorivna007:


src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+        .sort((a: Student, b: Student) => b[sortBy].localeCompare(a[sortBy]));
+
+    case SortType.Age:
+      return copyStudents
+        .sort((a: Student, b: Student) => b[sortBy] - a[sortBy]);
+
+    case SortType.AverageGrade:
+      return copyStudents
+        .sort((a: Student, b: Student) => averageGradeCalc(b.grades)
+          - averageGradeCalc(a.grades));
```
loralevitska, line 67: Good job. Let's improve. In order not to duplicate the code, move it to a helper function and. depending on how you need to sort, the order of the parameters will change, for example

```
if (order === 'asc') {
  return myHelperFunction(a, b)
} 

return myHelperFunction(b, a)
```

loralevitska (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+      case SortType.Surname:
+        return sign * (a[sortBy].localeCompare(b[sortBy]));
+
+      case SortType.AverageGrade:
+        return sign
+          * (averageGradeCalc(a.grades) - averageGradeCalc(b.grades));
+
+      case SortType.Age:
+      case SortType.Married:
+        return sign * (+a[sortBy] - +b[sortBy]);
```
loralevitska, line 45: avoid such expressions `- +`. Use [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) convert 
```suggestion
        return sign * (+a[sortBy] - Number(b[sortBy]));
```

loralevitska (approved):

PR #1103, opened 2023-02-20
Solution

Vasyl-Pavlenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+          return order === 'asc'
+            ? Number(a[sortBy]) - Number(b[sortBy])
+            : Number(b[sortBy]) - Number(a[sortBy]);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+        default:
+          return order === 'asc'
+            ? getAverageGrade(a) - getAverageGrade(b)
+            : getAverageGrade(b) - getAverageGrade(a);
```
etojeDenys, line 49: it would be better to move it into separate `case` and in `default` just throw the error

etojeDenys (changes requested):

OleziO (approved):

PR #1105, opened 2023-02-21
Update

petrolozynskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,38 @@
...
+  }
 
+  if (sortBy === SortType.Age && order === 'asc') {
+    return newobj.sort((a: Student, b: Student) => a.age - b.age);
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return [];
 }
```
Volodymyr-Melnyk-Fron10der, line 38:
```suggestion
}

const students = [
  {
    name: 'Diana',
    surname: 'Dorsey',
    age: 24,
    married: false,
    grades: [3, 3, 4, 5, 4, 3, 5, 5],
  },
  {
    name: 'Christina',
    surname: 'Branscome',
    age: 23,
    married: true,
    grades: [4, 4, 4, 5, 5, 5, 5, 5],
  },
  {
    name: 'Willie',
    surname: 'Barrera',
    age: 22,
    married: false,
    grades: [3, 5, 5, 3, 3, 5, 4, 4],
  },
];

sortStudents(students )

```

src/sortStudents.ts
```diff
@@ -1,16 +1,38 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'desc' | 'asc';
+
+export function
+sortStudents(students: Student[], sortBy: SortType, order: SortOrder)
+  : Student[] |void {
+  const newobj = JSON.parse(JSON.stringify(students));
```
TarasHoliuk, line 27: you don't do any changes inside `Student` objects, so no need to do a deep copy

src/sortStudents.ts
```diff
@@ -1,16 +1,38 @@
...
+  averageGrade
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'desc' | 'asc';
+
+export function
+sortStudents(students: Student[], sortBy: SortType, order: SortOrder)
+  : Student[] |void {
```
TarasHoliuk, line 26:
```suggestion
export function sortStudents(students: Student[], sortBy: SortType, order: SortOrder): Student[] {
```
there shouldn't be `void` return type

TarasHoliuk (changes requested):
Please, finish the task and re-request a review after that
If you need help or have any questions - feel free to ask in chat

src/sortStudents.ts
Viktor-Kost: is not fixed

Viktor-Kost (changes requested):
Hi!)
I see you've not fixed previous mentor's remarks. So, if you need any help - feel free to ask at fe_chat

petrolozynskyi:
> Hi!) I see you've not fixed previous mentor's remarks. So, if you need any help - feel free to ask at fe_chat
"you don't do any changes inside Student objects, so there is no need to do a deep copy"

Hi, I don't fix this one remark because I think
We need to do a deep copy because the method "sort" is changing our array order. And in the test, we can see  "should not modify the original array"
The sort() method sorts the elements of an array [in place](https://en.wikipedia.org/wiki/In-place_algorithm). - This means, the changed order in Array, right?
In my code, I don't need to change the order of getting the array, so I do it like this. Why this remark?

SanyaBratashchuk:
> > Hi!) I see you've not fixed previous mentor's remarks. So, if you need any help - feel free to ask at fe_chat
> > "you don't do any changes inside Student objects, so there is no need to do a deep copy"
> 
> Hi, I don't fix this one remark because I think We need to do a deep copy because the method "sort" is changing our array order. And in the test, we can see "should not modify the original array" The sort() method sorts the elements of an array [in place](https://en.wikipedia.org/wiki/In-place_algorithm). - This means, the changed order in Array, right? In my code, I don't need to change the order of getting the array, so I do it like this. Why is this a mistake?

You only need a shallow copy here, because you don't change the elements of the array, only their order in it.

SanyaBratashchuk (changes requested):
Read my answer above. If you have any questions feel free to ask in Slack

petrolozynskyi:
> Read my answer above. If you have any questions feel free to ask in Slack

Got it, Done,

Volodymyr-Melnyk-Fron10der (changes requested):
I suggest to move all if statements which have common `order `value under one common condition. If you have any questions feel free to aske them in the Slack

petrolozynskyi:
> Done

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+        - Sum(a.grades) / a.grades.length);
+    }
+
+    if (sortBy === SortType.Married) {
+      return newobj.sort((
+        a:Student,
+        b:Student,
+      ) => Number(b.married) - Number(a.married));
+    }
+  }
```
anstsot, line 53: please add sorting by SortType.Name and SortType.Surname

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+        b:Student) => a.name.localeCompare(b.name));
+    }
 
+    if (sortBy === SortType.Surname) {
+      return newobj.sort((
+        a:Student,
+        b:Student,
+      ) => a.surname.localeCompare(b.surname));
+    }
+  }
```
anstsot, line 79: please add here sorting by SortType.Married

anstsot (changes requested):
Looks good in general but some sortings are missing

src/sortStudents.ts
```diff
@@ -1,16 +1,101 @@
...
+      ) => Number(b.married) - Number(a.married));
+    }
+
+    if (sortBy === SortType.Name) {
+      return newobj.sort((a:Student,
+        b:Student) => b.name.localeCompare(a.name));
+    }
+
+    if (sortBy === SortType.Surname) {
+      return newobj.sort((
```
Volodymyr-Melnyk-Fron10der, line 60: use switch/case here

src/sortStudents.ts
```diff
@@ -1,16 +1,101 @@
...
+        a:Student,
+        b:Student,
+      ) => a.surname.localeCompare(b.surname));
+    }
 
+    if (sortBy === SortType.Married) {
+      return newobj.sort((
+        a:Student,
```
Volodymyr-Melnyk-Fron10der, line 94: here as well

Volodymyr-Melnyk-Fron10der (changes requested):
almost done👍

PR #1109, opened 2023-02-23
solution

anna-akhalmosulishvili:


src/sortStudents.ts
```diff
@@ -1,16 +1,85 @@
...
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  switch (order) {
+    case 'asc':
+      return sortAsc(studentsCopy, sortBy);
```
IvanFesenko, line 78: Lets rework and make you code shorter

use swith to sortBy, then use ternary operator for sortOrder. You will have something like this
```
    switch (sortBy) {
      case SortType.Name:
        return order === 'asc'
```

IvanFesenko (changes requested):
Good job, let make some improvements

ViktorHavryliuk (approved):

PR #1110, opened 2023-02-23
Solution

stanislavcodes:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const calcAverageGrade = ({ grades }: Student): number => (
+    grades.reduce((total, grade) => total + grade, 0) / grades.length
+  );
```
OleksiiBlezniuk, line 26: I guess we shouldn't create this function every time when `sortStudents` is executed, let's move it outside

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+
+  const studentsCopy: Student[] = [...students];
```
OleksiiBlezniuk, line 28: No need to specify type gere

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  const isAscendingOrder = order === 'asc';
```
OleksiiBlezniuk, line 29: A bit overkill as for me, but ok

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+          ? calcAverageGrade(firstStudent) - calcAverageGrade(secondStudent)
+          : calcAverageGrade(secondStudent) - calcAverageGrade(firstStudent);
+      case SortType.Age:
+        return isAscendingOrder
+          ? firstStudent[sortBy] - secondStudent[sortBy]
+          : secondStudent[sortBy] - firstStudent[sortBy];
+      case SortType.Married:
+        return isAscendingOrder
+          ? Number(firstStudent[sortBy]) - Number(secondStudent[sortBy])
+          : Number(secondStudent[sortBy]) - Number(firstStudent[sortBy]);
```
OleksiiBlezniuk, line 49: Is these conditions logic different? Let's combine these cases

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        return isAscendingOrder
+          ? firstStudent[sortBy].localeCompare(secondStudent[sortBy])
+          : secondStudent[sortBy].localeCompare(firstStudent[sortBy]);
+      case SortType.AverageGrade:
+        return isAscendingOrder
+          ? calcAverageGrade(firstStudent) - calcAverageGrade(secondStudent)
+          : calcAverageGrade(secondStudent) - calcAverageGrade(firstStudent);
```
OleksiiBlezniuk, line 41:
```suggestion
      case SortType.Surname:
        return isAscendingOrder
          ? firstStudent[sortBy].localeCompare(secondStudent[sortBy])
          : secondStudent[sortBy].localeCompare(firstStudent[sortBy]);
          
      case SortType.AverageGrade:
        return isAscendingOrder
          ? calcAverageGrade(firstStudent) - calcAverageGrade(secondStudent)
          : calcAverageGrade(secondStudent) - calcAverageGrade(firstStudent);
```

OleksiiBlezniuk: This way will be more readable, separate big blocks of code by blank lines

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  order: SortOrder,
+): Student[] {
+  const calcAverageGrade = ({ grades }: Student): number => (
+    grades.reduce((total, grade) => total + grade, 0) / grades.length
+  );
+
+  const studentsCopy: Student[] = [...students];
+  const isAscendingOrder = order === 'asc';
 
+  studentsCopy.sort((firstStudent, secondStudent): number => {
```
OleksiiBlezniuk, line 31: Or you can return this `sort` result

OleksiiBlezniuk (changes requested):

kuznetsova-anastasiia (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  return studentsCopy.sort((firstStudent, secondStudent): number => {
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'asc'
```
illiarerun, line 34: cause we could've used here something like
```suggestion
        return order === SortType.Ascending
```
and to be dependant on plain strings

illiarerun (approved):

PR #1111, opened 2023-02-24
solution

ivanm2706:


src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+type Element = string | number | boolean | number[];
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const arr: Student[] = [...students];
```
VitaliyBondarenko1982, line 25: better naming copyStudents, cloneStudents

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+    let left: Element = a[sortBy];
+    let right: Element = b[sortBy];
+    const o: Element = left;
+
+    if (order === 'desc') {
+      left = right;
+      right = o;
+    }
+
+    switch (typeof left) {
```
VitaliyBondarenko1982, line 37: Suggesting to write switch which depends on sortBy It wiill be more understandable and we will no need additional type Element

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const arr: Student[] = [...students];
+
+  return arr.sort((a: Student, b: Student): number => {
+    let left: Element = a[sortBy];
+    let right: Element = b[sortBy];
+    const o: Element = left;
```
VitaliyBondarenko1982, line 30: Try to think about better naming instead o

VitaliyBondarenko1982 (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const cloneStudent: Student[] = [...students];
+
+  return cloneStudent.sort((a: Student, b: Student): number => {
+    let left: Element = a[sortBy];
+    let right: Element = b[sortBy];
```
IvanFesenko, line 29: What is left and right? Hand or leg? Please fix naming was mentioned above

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+      case 'surname':
+        return left.localeCompare(right);
+
+      case 'age':
+        return left - right;
+
+      case 'married':
+        return left - right;
 
+      case 'grades':
```
IvanFesenko, line 50: You have SortType enum for it, do not use string

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+        return left.reduce((
+          accLeft: number,
+          currLeft: number,
+        ): number => accLeft + currLeft)
+        / left.length
+        - right.reduce((
+          accRight: number,
+          currRight: number,
+        ): number => accRight + currRight)
```
IvanFesenko, line 59: use `Array.sort()`

IvanFesenko (changes requested):
Needs refactoring. If you have any questions - ask them in fe_chart

src/sortStudents.ts
ivanm2706: don`t understand why I must use sort()?

mykhalenych (approved):

PR #1121, opened 2023-03-01
● Solution

d-hubych:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = JSON.parse(JSON.stringify(students));
```
SerhiiKirik, line 25: why do you need this?)
```suggestion
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  const reducer = (acc: number, curt: number): number => acc + curt;
```
SerhiiKirik, line 26:
```suggestion
  const callback = (acc: number, curt: number): number => acc + curt;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  sortedStudents.sort((a: Student, b: Student): number => {
+    let comparator: number = 0;
+
+    switch (sortBy) {
+      case ('name'):
```
SerhiiKirik, line 32: use Enum for this
in all cases
```suggestion
      case (SortType.Name):
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = JSON.parse(JSON.stringify(students));
+  const reducer = (acc: number, curt: number): number => acc + curt;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  sortedStudents.sort((a: Student, b: Student): number => {
```
SerhiiKirik, line 28:
```suggestion
  sortedStudents.sort((studentA: Student, studentB: Student): number => {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+      case ('age'):
+      case ('married'):
+        comparator = +a[sortBy] - +b[sortBy];
+        break;
+      case ('grades'):
+        comparator = (a[sortBy].reduce(reducer) / a[sortBy].length)
+          - (b[sortBy].reduce(reducer) / b[sortBy].length);
+        break;
+      default:
+        break;
```
SerhiiKirik, line 45:
```suggestion
        break;
        
     default:
     throw new Error('What is happening?(9');
```

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+    let comparator: number = 0;
+
+    switch (sortBy) {
+      case ('name'):
+      case ('surname'):
+        comparator = a[sortBy].localeCompare(b[sortBy]);
+        break;
+      case ('age'):
+      case ('married'):
+        comparator = +a[sortBy] - +b[sortBy];
```
SerhiiKirik, line 38: can you do this without comparator?

```suggestion
        return +a[sortBy] - +b[sortBy]
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's improve you code 🧑‍💻

src/sortStudents.ts
tetiana-ivanchenko, line 32:
```suggestion
      case SortType.Name:
```

src/sortStudents.ts
tetiana-ivanchenko, line 25:
```suggestion
  const copyStudents = [...students];
```

tetiana-ivanchenko (changes requested):

src/sortStudents.ts
d-hubych: I can, but now I do this once `return (order === 'asc') ? comparator : comparator * -1;`
and without `comparator` I'll need to do this in each case
`return (order === 'asc') ? (+a[sortBy] - +b[sortBy]) : (+a[sortBy] - +b[sortBy]) * -1;`

So what is better?

SerhiiKirik: Let's think
Why should the code continue if we have already performed the action we need?
Why should we go through further checks?

etojeDenys (approved):
looks awesome 🤩

PR #1124, opened 2023-03-03
Solution

RuslanLiapin:


src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+
+            if (b.married) {
+              return 1;
+            }
+
+            return -1;
+
+          default:
+            return -1;
+        }
```
loralevitska, line 63: Use helper function for logic and then revert params. Ex. if asc call function(a, b) else call function(b, a)

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

VitaliyBondarenko1982 (approved):

PR #1125, opened 2023-03-04
Solution

timoxa2t:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  const sortCallback : Callback = (first, second) => {
+    let difrence: number;
+
+    switch (sortBy) {
+      case SortType.Name:
+        difrence = first.name < second.name ? -1 : 1;
+        break;
+      case SortType.Surname:
+        difrence = first.surname < second.surname ? -1 : 1;
+        break;
```
kirill-vypirovskyi, line 46: You can set values in enum like
Name = 'name'
and then here you dont need to compare name and surna,e separately
just first[sortBy] and second[sortBy]
sortBy will dinamically changed and you don't need to repeat same logic.

And there is a better method to compare strings - localeCompare

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+      case SortType.Age:
+        difrence = first.age - second.age;
+        break;
+      case SortType.Married:
+        difrence = Number(first.married) - Number(second.married);
```
kirill-vypirovskyi, line 51: The same here, u can do it in one, just with Number(first[sortBy] - Number(second[sortBy]

kirill-vypirovskyi (approved):
It will be better to throw an error when in default case, because it is impossible to get unknown sort type.

And instead of break evety time you can just return value from switch

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
+/* eslint-disable comma-dangle */
```
sergii-nosachenko, line 1: Don't disable linter

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder
+): Student[] {
+  type Callback = (first: Student, second: Student) => number;
+
+  const sortCallback : Callback = (first, second) => {
+    let difrence: number;
```
sergii-nosachenko, line 38:
```suggestion
    let difference: number;
```

src/sortStudents.ts
sergii-nosachenko: Good point. @timoxa2t fix this

src/sortStudents.ts
sergii-nosachenko: Same here

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+        difrence = Number(first.married) - Number(second.married);
+        break;
+      case SortType.AverageGrade:
+        difrence = getAvgGrade(first) - getAvgGrade(second);
+        break;
+      default:
+        difrence = 0;
+    }
+
+    return order === 'asc' ? difrence : -difrence;
```
sergii-nosachenko, line 60: 🔥

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder
+): Student[] {
+  type Callback = (first: Student, second: Student) => number;
+
+  const sortCallback : Callback = (first, second) => {
```
sergii-nosachenko, line 37:
```suggestion
  const sortCallback: Callback = (first, second) => {
```

src/sortStudents.ts
sergii-nosachenko, line 46: Cases can be combined

src/sortStudents.ts
sergii-nosachenko, line 51: Cases can be combined

sergii-nosachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
 }
 
 // create SortOrder type
```
sergii-nosachenko, line 18: Cleanup code before PR

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+        difrence = 0;
+    }
+
+    return order === 'asc' ? difrence : -difrence;
+  };
+  const studentsCopy: Student[] = [...students];
+
+  studentsCopy.sort(sortCallback);
+
+  return studentsCopy;
```
sergii-nosachenko, line 60:
```suggestion
  return [...students].sort(sortCallback);
```
Can be optimized.

sergii-nosachenko (changes requested):
Not all previous comments are fixed.

sergii-nosachenko (approved):
Good work 👍

PR #1126, opened 2023-03-04
solution

VovaKlimyuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+
+    case 'object': {
+      const s1Avg = (s1[sortBy] as number[])
+        .reduce((acc, curr) => acc + curr, 0) / (s1[sortBy] as number[])
+        .length;
+      const s2Avg = (s2[sortBy] as number[])
+        .reduce((acc, curr) => acc + curr, 0) / (s2[sortBy] as number[])
+        .length;
+
+      return order === 'asc' ? s1Avg - s2Avg : s2Avg - s1Avg;
```
vkryvytskyy, line 54: You're checking order in every case, think about optimising this

vkryvytskyy (approved):

PR #1128, opened 2023-03-05
Develop

Zelinskyi-Serhii:


src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+      if (order === 'asc') {
+        copyObject.sort((prev, current) => {
+          if (current.married === prev.married) {
+            return 0;
+          }
+
+          return current.married ? -1 : 1;
+        });
+      }
+      break;
```
vladyslav-mosor, line 71: You could use the same logic as for sorting by number type instead of this. Because your boolean values could be converted to number values.

src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+        }, 0);
+        const bSum = current.grades.reduce((acc: number, num: number) => {
+          return acc + num;
+        }, 0);
+
+        return order === 'asc'
+          ? aSum / prev.grades.length - bSum / current.grades.length
+          : bSum / current.grades.length - aSum / prev.grades.length;
+      });
+      break;
```
vladyslav-mosor, line 86: Try to create a helper function for calculating average grade so you dont have to repeat your code.

vladyslav-mosor (changes requested):
Cheklist p.3. Try to rewrite your switch-case construction so you cas use the same logic for sorting by name or surname (because they both have the type string so you dont have to repeat your code). The same for sorting by age and married.

+ if you use return inside switch construction it`s not necessary to use break.

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' |'desc';
```
sergii-nosachenko, line 18:
```suggestion
export type SortOrder = 'asc' | 'desc';
```

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
 
+function countSum(numbers: number[]): number {
```
sergii-nosachenko, line 20: Better refactor this helper function to calculate the average value.

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  return numbers.reduce((acc: number, num: number) => {
+    return acc + num;
+  }, 0);
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const copyArray = [...students];
+  const copyObject = copyArray.map((person) => ({ ...person }));
```
sergii-nosachenko, line 30: You don't need to make a deep copy of the array. You are changing only the order of objects, objects themselves are staying untouched.

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyObject.sort((prev, current) => {
+        return order === 'asc'
+          ? prev[sortBy].localeCompare(current[sortBy])
+          : current[sortBy].localeCompare(prev[sortBy]);
```
sergii-nosachenko, line 38: DRY: you don't need to have both comparisons. Just make one and multiply the result by -1 in the case of `desc` order.
Fix it everywhere.

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const copyArray = [...students];
+  const copyObject = copyArray.map((person) => ({ ...person }));
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyObject.sort((prev, current) => {
```
sergii-nosachenko, line 35:
```suggestion
      return copyObject.sort((prev, current) => {
```
You can return a result immediately.

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+        const bSum = countSum(current.grades);
+
+        return order === 'asc'
+          ? aSum / prev.grades.length - bSum / current.grades.length
+          : bSum / current.grades.length - aSum / prev.grades.length;
+      });
+      break;
+
+    default:
+      throw new Error('wrong parameters');
```
sergii-nosachenko, line 63: Provide detailed message. Use template string to show faulty `sortBy` value.

sergii-nosachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortCopyArray.sort((prev, current) => {
+        const compare = prev[sortBy].localeCompare(current[sortBy]);
+
+        return order === 'asc'
+          ? compare
+          : compare * -1;
```
OleksiiBlezniuk, line 43: Looks like this multiplier can be moved to constant, it's same for all cases

OleksiiBlezniuk (changes requested):
Let's simplify your code

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+  return sum / numbers.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortCopyArray = [...students];
+  const adscCompare = -1;
```
sergii-nosachenko, line 34:
```suggestion
  const direction = order === 'asc' ? 1 : -1;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortCopyArray.sort((prev, current) => {
+        const compare = prev[sortBy].localeCompare(current[sortBy]);
+
+        return order === 'asc'
+          ? compare
+          : compare * adscCompare;
```
sergii-nosachenko, line 44: You can immediately return a result from callback multiplied by `direction`.

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+        const bSum = countAverageSum(current.grades);
+        const compare = aSum - bSum;
+
+        return order === 'asc'
+          ? compare
+          : compare * adscCompare;
+      });
+
+    default:
+      throw new Error('Wrong sort type parameters');
```
sergii-nosachenko, line 69: Be more specific: use template string and sortBy value.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1130, opened 2023-03-06
Solution

Ukrainiane-panda:


src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+        break;
+
+      case SortType.Married:
+        if (a.married === b.married) {
+          compareValue = 0;
+        } else if (a.married) {
+          compareValue = 1;
+        } else {
+          compareValue = -1;
+        }
```
Volodymyr-Melnyk-Fron10der, line 50: it should works the same way
```suggestion
          compareValue = Number(a.married !== b.married);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,84 @@
...
+        break;
+
+      case SortType.AverageGrade:
+        avgGradeA = a.grades.reduce(
+          (sum: number, x: number) => (sum + x), 0,
+        ) / a.grades.length;
+
+        avgGradeB = b.grades.reduce(
+          (sum: number, x: number) => (sum + x), 0,
+        ) / b.grades.length;
```
Volodymyr-Melnyk-Fron10der, line 60: move duplicated code into function and reuse it

Volodymyr-Melnyk-Fron10der (changes requested):
Good job actually, just a few moments I see

PR #1135, opened 2023-03-07
Solution

nazarovoleksiy:


vadosik15 (approved):
Gj

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  return [...students].sort((previous, current) => {
+    switch (sortBy) {
+      case SortType.Surname:
+      case SortType.Name:
+        return order === 'asc'
+          ? previous[sortBy].localeCompare(current[sortBy])
+          : current[sortBy].localeCompare(previous[sortBy]);
```
sergii-nosachenko, line 35: DRY: you don't need to have both comparisons. Just make one and multiply the result by -1 in the case of the `desc` order.
Fix it everywhere.

sergii-nosachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+          * orderType;
+
+      case SortType.AverageGrade:
+        return (calcAverage(previous) - calcAverage(current))
+          * orderType;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        return 0;
```
sergii-nosachenko, line 48: Better throw an error with the reason of fault.

sergii-nosachenko (approved):
Good work 👍

PR #1137, opened 2023-03-07
solution

Nordexz:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? students.map((el) => el)
+          .sort((a, b) => Number(a[sortBy]) - Number(b[sortBy]))
+        : students.map((el) => el)
+          .sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
```
MarkBabich, line 40: As for me its better to make a copy of `students` array once at the start of the function, and than just mutate it with `sort`

MarkBabich (changes requested):
Some fixes, and task is done :)

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? [...students]
+          .sort((firstStudent, secondStudent) => firstStudent[sortBy]
+            .localeCompare(secondStudent[sortBy]))
+        : [...students]
+          .sort((firstStudent, secondStudent) => secondStudent[sortBy]
+            .localeCompare(firstStudent[sortBy]));
```
OleksiiBlezniuk, line 34: You've written it in all cases
Let's follow DRY and move switch/case inside sort method

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+          .sort((firstStudent, secondStudent) => Number(secondStudent[sortBy])
+            - Number(firstStudent[sortBy]));
+
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? [...students]
+          .sort((firstStudent, secondStudent) => (firstStudent.grades
+            .reduce((acc, num) => acc + num, 0) / firstStudent.grades.length)
+              - (secondStudent.grades
+                .reduce((acc, num) => acc + num)) / secondStudent.grades.length)
```
OleksiiBlezniuk, line 52: Let's create helper function for calculating average value

OleksiiBlezniuk (changes requested):

Andr1yK (approved):

PR #1144, opened 2023-03-07
TS Sort Students - Solution

Yevhenii0536:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' |'desc';
 
+function getAverageGrade(grades: number[]): number {
```
KuharchukDmytro, line 20: you can use destructuring, so when you call the function you can just pass an object and it will automaticly extract grades
```suggestion
function getAverageGrade({ grades }: Student): number {
```

KuharchukDmytro (approved):

OleksiiBlezniuk (approved):

PR #1146, opened 2023-03-07
sortStudents solution

kateryna-babina:


src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+      case SortType.Age:
+      case SortType.Married:
+        return order === 'asc'
+          ? Number(firstStudent[sortBy]) - Number(secondStudent[sortBy])
+          : Number(secondStudent[sortBy]) - Number(firstStudent[sortBy]);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        return 0;
```
MarkBabich, line 51: Default case should throw an error

MarkBabich (approved):
GJ :)

src/sortStudents.ts
sergii-nosachenko: Good point. Needs to be fixed.

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+): Student[] {
+  const sortedStudents = [...students];
+
+  return sortedStudents.sort((firstStudent, secondStudent) : number => {
+    switch (sortBy) {
+      case (SortType.Name):
+      case (SortType.Surname):
+        return order === 'asc'
+          ? firstStudent[sortBy].localeCompare(secondStudent[sortBy])
+          : secondStudent[sortBy].localeCompare(firstStudent[sortBy]);
```
sergii-nosachenko, line 37: DRY: you don't need to have both comparisons. Just make one and multiply the result by -1 in the case of the `desc` order.
Fix it everywhere.

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  return sortedStudents.sort((firstStudent, secondStudent) : number => {
```
sergii-nosachenko, line 31:
```suggestion
  return [...students].sort((firstStudent, secondStudent) : number => {
```
Unnecessary variable.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1148, opened 2023-03-07
solution

Lilu580:


milenadiamonds1 (changes requested):
GJ but need some fixes

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+type AverageGrade = (x : Student) => number;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+const findSummOfGrades = (
+  prev : number, grade: number,
+): number => prev + grade;
+
+const getAverageGrade: AverageGrade
+= (object) => object.grades.reduce(findSummOfGrades, 0) / object.grades.length;
```
OleksiiBlezniuk, line 27:
```suggestion
const getAverageGrade: AverageGrade = (object) => (
  object.grades.reduce(
    findSummOfGrades,
    0
  ) / object.grades.length
)
```

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+
+export function sortStudents(
+  students : Student[], sortBy : SortType, order : SortOrder,
+) : Student[] {
+  const arrayOfStudents = [...students];
+  const checkOrder = order === 'asc';
```
OleksiiBlezniuk, line 33: I think it's an overkill here

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return checkOrder
+          ? prev[sortBy].localeCompare(curr[sortBy])
+          : curr[sortBy].localeCompare(prev[sortBy]);
+      case SortType.Age:
+      case SortType.Married:
+        return checkOrder
+          ? (+prev[sortBy]) - (+curr[sortBy])
+          : (+curr[sortBy]) - (+prev[sortBy]);
```
OleksiiBlezniuk, line 46: Use `Number` constructor instead of `+`

OleksiiBlezniuk (approved):

PR #1156, opened 2023-03-08
solution

chernenko2:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const arrayOfStudents = [...students];
+
+  return arrayOfStudents.sort((student1: Student, student2: Student) => {
```
Stepan-TS, line 34: I think it is better when the parameter and type are in the same line)

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        return order === SortOrder.Asc
+          ? student1[sortBy].localeCompare(student2[sortBy])
+          : student2[sortBy].localeCompare(student1[sortBy]);
```
Stepan-TS, line 40: you can use only one condition and multiply the result  by -1 in the case of the desc order)

Stepan-TS (approved):

src/sortStudents.ts
sergii-nosachenko, line 34: Useless variable. Just immediately return a sorted copy.

src/sortStudents.ts
sergii-nosachenko: Good point. Fix this.

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+          : Number(student2[sortBy]) - Number(student1[sortBy]);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      case SortType.AverageGrade:
+        return order === SortOrder.Asc
+          ? averageGrades(student1) - averageGrades(student2)
+          : averageGrades(student2) - averageGrades(student1);
+      default:
+        throw new Error('It isn`t a valid type');
```
sergii-nosachenko, line 53:
```suggestion
        throw new Error(`${sortBy} isn't a valid sorting type`);
```
Describe reason.

sergii-nosachenko (changes requested):

Andr1yK (changes requested):

sergii-nosachenko (approved):

PR #1158, opened 2023-03-08
Task complete

andriibezkorovainyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsOut = students.map((student) => (
```
OleksiiBlezniuk, line 25: What does studentsOut means?

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+    {
+      ...student,
+      averageGrade: student.grades.reduce(
+        (prevGrade, currGrade) => prevGrade + currGrade,
+      ) / student.grades.length,
+    }
+  ));
```
OleksiiBlezniuk, line 32: But why should we map our array and add additional property to student if `sortBy` is not equal to averageGrade?

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+      case SortType.Age:
+      case SortType.AverageGrade:
+        return order === 'desc'
+          ? bStudent[sortBy] - aStudent[sortBy]
+          : aStudent[sortBy] - bStudent[sortBy];
+
+      case SortType.Married:
+        return order === 'desc'
+          ? Number(bStudent.married) - Number(aStudent.married)
+          : Number(aStudent.married) - Number(bStudent.married);
```
OleksiiBlezniuk, line 51: Btw, these cases looks pretty similar, let's combine them

OleksiiBlezniuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+    (prevGrade, currGrade) => prevGrade + currGrade,
+  ) / student.grades.length;
+
+  return [...students].sort((aStudent: Student, bStudent: Student): number => {
+    switch (sortBy) {
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'desc'
+          ? bStudent[sortBy].localeCompare(aStudent[sortBy])
+          : aStudent[sortBy].localeCompare(bStudent[sortBy]);
```
sergii-nosachenko, line 34: DRY: you don't need to have both comparisons. Just make one and multiply the result by -1 in the case of the `desc` order.
Fix it everywhere.

src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
 
+      case SortType.AverageGrade:
+        return order === 'desc'
+          ? getAverageGrade(bStudent) - getAverageGrade(aStudent)
+          : getAverageGrade(aStudent) - getAverageGrade(bStudent);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        return 0;
```
sergii-nosachenko, line 48: Throw an error and describe the reason. Use template string to display faulty `sortBy` value.

sergii-nosachenko (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const getAverageGrade = (student: Student): number => student.grades.reduce(
+    (prevGrade, currGrade) => prevGrade + currGrade,
+  ) / student.grades.length;
```
OleksiiBlezniuk, line 26: Move helper functions outside main function

src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
+  const orderDirection = order === 'asc' ? 1 : -1;
+
+  return [...students].sort((aStudent: Student, bStudent: Student): number => {
+
+    switch (sortBy) {
```
OleksiiBlezniuk, line 31:
```suggestion
  return [...students].sort((aStudent: Student, bStudent: Student): number => {
    switch (sortBy) {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,50 @@
...
+          * orderDirection;
 
+      case SortType.AverageGrade:
+        return (getAverageGrade(aStudent) - getAverageGrade(bStudent))
+          * orderDirection;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        throw new Error(`Invalid sort criteria: "${sortBy}"`);
```
OleksiiBlezniuk, line 47: I think error is an overkill here, just return 0

OleksiiBlezniuk (approved):

PR #1159, opened 2023-03-08
sortStudent

Ambulence:


Vlados3611 (approved):
I see no reason to put a reject :) Good Job!

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+        * sort;
 
+      case SortType.AverageGrade:
+        return (getAverage(firstPupil) - getAverage(secondPupil))
+        * sort;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        throw new Error();
```
sergii-nosachenko, line 53: Throw an error and describe the reason. Use template string to display faulty `sortBy` value.

sergii-nosachenko (changes requested):
Quick fix is needed

OleksiiBlezniuk (approved):

PR #1168, opened 2023-03-09
Solution

arkushenko-anna:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+        return (+prevStudent[sortBy] - +curStudent[sortBy]) * sortingOrder;
+
+      case SortType.AverageGrade:
+        return (getAverageGrade(prevStudent) - getAverageGrade(curStudent))
+        * sortingOrder;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      default:
+        throw new Error('Type is not valid');
```
OleksiiBlezniuk, line 52: I think it's an overkill here, just return 0

OleksiiBlezniuk (approved):

PR #1172, opened 2023-03-10
task completed

ArthurKrepko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+          ? calculateAverageGrade(prev) - calculateAverageGrade(curr)
+          : calculateAverageGrade(curr) - calculateAverageGrade(prev);
+      case SortType.Age:
+        return order === 'asc'
+          ? +prev[sortBy] - +curr[sortBy]
+          : +curr[sortBy] - +prev[sortBy];
+      case SortType.Married:
+        return order === 'asc'
+          ? +prev[sortBy] - +curr[sortBy]
+          : +curr[sortBy] - +prev[sortBy];
```
OleksiiBlezniuk, line 42: These cases are same, better to combine them instead of repeating, follow DRY

OleksiiBlezniuk: And better to use `Number` constructor instead of `+`, it's more readable

OleksiiBlezniuk (approved):
Approve, but pay attention to comments

PR #1174, opened 2023-03-10
the task was solved

batiukova-nataliia:


etojeDenys (changes requested):
let`s rewrite it using switch\case

VitaliyBondarenko1982 (approved):

PR #1175, opened 2023-03-11
sort_solution

oleksiy-uchaev:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
 
 export interface Student {
-  // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: [],
```
TarasHoliuk, line 7: it's better to set more specified array type. i.e `number[]`

TarasHoliuk: FYI

TarasHoliuk (approved):
LGTM :+1:

PR #1181, opened 2023-03-14
Solution

OlhaKissa:


src/sortStudents.ts
```diff
@@ -1,16 +1,41 @@
...
+      return a[sortBy].localeCompare(b[sortBy]);
+    }
+
+    if (sortBy === 'age') {
+      return a[sortBy] - b[sortBy];
+    }
+
+    if (sortBy === 'married') {
+      return a[sortBy] - b[sortBy];
+    }
```
IvanFesenko, line 37: Simplify using ||

src/sortStudents.ts
```diff
@@ -1,16 +1,41 @@
...
+  });
+
+  return sortedStudents;
```
IvanFesenko, line 40:
```suggestion

```

IvanFesenko (changes requested):
Check my comments. Also check lint errors

etojeDenys (changes requested):
fixed these errors

src/sortStudents.ts
```diff
@@ -1,16 +1,35 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+): Student[] {
+  const sortedStudents = JSON.parse(JSON.stringify(students));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return sortedStudents.sort((a: Student, b: Student): number => {
+    if (sortBy === 'name' || sortBy === 'surname') {
```
etojeDenys, line 27: you need to use here SortType enum

src/sortStudents.ts
```diff
@@ -1,16 +1,35 @@
...
+      return a[sortBy].localeCompare(b[sortBy]);
+    }
+
+    if (sortBy === 'age' || sortBy === 'married') {
+      return a[sortBy] - b[sortBy];
```
etojeDenys, line 32: and it will better it you rewrite it with switch\case

etojeDenys (changes requested):
also I don't see any logic with asc || desc & sorting by grades

OlEzhA0 (approved):

PR #1185, opened 2023-03-14
Solution-1

MykolaButylkov:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+  let resultArray : Student[] = [...students];
+
+  function callbackGrades(
+    previousValue: number, currentValue: number,
+  ) : number {
+    return previousValue + currentValue;
+  }
+
+  switch (sortBy) {
+    case 'age':
```
etojeDenys, line 36: we can use here enum instead of string

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+      resultArray = resultArray.sort((a, b) => +a.age - +b.age);
```
etojeDenys, line 37: `.sort()` is not immutable
```suggestion
      resultArray.sort((a, b) => +a.age - +b.age);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+      resultArray = [...resultArray].sort((a, b) => (
+        a.grades.reduce(callbackGrades) / a.grades.length
+      ) - (b.grades.reduce(callbackGrades) / b.grades.length));
+      break;
+
+    default:
+      break;
+  }
+
+  return order === 'desc' ? resultArray.reverse() : resultArray;
```
etojeDenys, line 60: same here, use enum

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    return previousValue + currentValue;
+  }
+
+  switch (sortBy) {
+    case 'age':
+      resultArray = resultArray.sort((a, b) => +a.age - +b.age);
+      break;
+
+    case 'married':
+      break;
```
etojeDenys, line 41: you need to implement sorting by SortType.Married

etojeDenys (changes requested):
and look at tests, I see some problems there

OlEzhA0 (approved):

PR #1192, opened 2023-03-19
Solution

yana-kotsulym:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+  AverageGrade = 'AverageGrade',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
```
DarkMistyRoom, line 23:
```suggestion
export function sortStudents(
  students: Student[], 
  sortBy: SortType, 
  order: SortOrder,
): Student[] {
```

DarkMistyRoom: if you brake the line, place one argument per line

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        return order === 'asc'
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
+      });
+      break;
+    case SortType.AverageGrade:
+      copiedStudents.sort((a: Student, b: Student) => {
+        const curr: number
+          = a.grades.reduce((sum: number, el: number) => sum + el, 0)
+          / a.grades.length;
```
DarkMistyRoom, line 46: would be great to move this logic to a separate function

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      break;
+
+    case SortType.Married:
+      copiedStudents.sort((a: Student, b: Student) => {
+        const curr: number = a.married === true ? 1 : 0;
+        const next: number = b.married === true ? 1 : 0;
+
+        return order === 'asc'
+          ? curr - next
+          : next - curr;
```
DarkMistyRoom, line 64:
```suggestion
        return order === 'asc'
          ? a.married - b.married
          : b.married - a.married;
```

DarkMistyRoom: it will work because of type conversions

DarkMistyRoom (changes requested):
It's a good job. take just a few suggestions.
You can reduce the code by using `flag: 1 | -1` variable for specifying the sort order
`arr.sort((a, b) => callback(a, b) * flag);`

PR #1201, opened 2023-03-22
solution

MSA1991:


src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
 }
 
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudent = [...students];
```
```suggestion
  const copyOfStudents = [...students];
```
Actually, they are not sorted right now. It's just a copy of students

polosanya (changes requested):
Pay attention to the tests. Make sure that you sort students correctly
image

Volodymyr-Melnyk-Fron10der (approved):
Ok👍

PR #1205, opened 2023-03-27
Solution

bojkovladislav:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+    case SortType.AverageGrade:
+      sortedStudents.sort((student1, student2) => {
+        return order === "asc"
+          ? getAverageGrade(student1[sortBy]) -
+              getAverageGrade(student2[sortBy])
+          : getAverageGrade(student2[sortBy]) -
+              getAverageGrade(student1[sortBy]);
+      });
+      break;
+  }
```
Volodymyr-Melnyk-Fron10der, line 53: always add default case

Volodymyr-Melnyk-Fron10der (changes requested):
amazing, just one point to check

vlad-demchuk (approved):

PR #1207, opened 2023-03-28
Solution

vlad-ptashnichenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    case SortType.Name:
+      copyStudents.sort((a, b) => (order === 'asc'
+        ? a.name.localeCompare(b.name)
+        : b.name.localeCompare(a.name)));
+      break;
+    case SortType.Surname:
+      copyStudents.sort((a, b) => (order === 'asc'
+        ? a.surname.localeCompare(b.surname)
+        : b.surname.localeCompare(a.surname)));
+      break;
```
OlEzhA0, line 37:
```suggestion
    case SortType.Name:
    case SortType.Surname:
      copyStudents.sort((a, b) => (order === 'asc'
        ? a[sortBy].localeCompare(b[sortBy])
        : b[sortBy].localeCompare(a[sortBy])));
      break;
```

OlEzhA0 (changes requested):
hey! I added small improvement; could you check it pls
also apply it to similar places

PR #1210, opened 2023-03-28
created sort students function

prokivskaaa:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      return firstStudent[sortBy].toString()
+        .localeCompare(secondStudent[sortBy].toString());
+    }
+
+    if (sortBy === 'grades') {
+      const averageFirst: number = firstStudent.grades
+        .reduce((acc, el) => acc + el, 0) / firstStudent.grades.length;
+
+      const averageSecond: number = secondStudent.grades
+        .reduce((acc, el) => acc + el, 0) / secondStudent.grades.length;
```
jstmpelowycz, line 40: looks like there could be a dedicated function calculating average

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      if (order === 'desc') {
+        return secondStudent[sortBy].toString()
+          .localeCompare(firstStudent[sortBy].toString());
+      }
+
+      return firstStudent[sortBy].toString()
+        .localeCompare(secondStudent[sortBy].toString());
+    }
+
+    if (sortBy === 'grades') {
```
jstmpelowycz, line 35: instead of if-statements, you can try using switch-case
```js
switch (sortBy) {
  SortType.Name:
  SortType.Surname:
    // sort strings
  ...
}
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+
+      return +averageFirst - +averageSecond;
+    }
+
+    if (order === 'desc') {
+      return +secondStudent[sortBy] - +firstStudent[sortBy];
+    }
+
+    return +firstStudent[sortBy] - +secondStudent[sortBy];
+  });
```
jstmpelowycz, line 54: here's a trick you can use, you just swap them
```js
const firstStudentRef = firstStudent;
const secondStudentRef = secondStudent;

if (order === 'desc') {
  [firstStudentRef, secondStudentRef] = [secondStudentRef, firstStudentRef];
}
```

jstmpelowycz (approved):

PR #1211, opened 2023-03-29
solution :)

katerynashylina:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function findAverage(arr: number[]): number {
+  return arr.reduce((acc, curr) => acc + curr, 0) / arr.length;
```
Volodymyr-Melnyk-Fron10der, line 23: you can save 1 iteration here) if you do not set default value then it starts from the 1st element in array right away
```suggestion
  return arr.reduce((acc, curr) => acc + curr) / arr.length;
```

Volodymyr-Melnyk-Fron10der (approved):
Well done🔥

PR #1220, opened 2023-04-03
solution

Sbavdik:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  return grades
+    .reduce((acc, curr) => acc + curr, 0)
+    / grades.length;
+};
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: string,
+): Student[] {
```
artem-kyrylenko, line 29: i think would be better to specify that its not just a string but type SortOrder

artem-kyrylenko (approved):
Well done!

src/sortStudents.ts
liliia-livandovskaya: Agree

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAverageGrades(a[sortBy]) - getAverageGrades(b[sortBy])
+          : getAverageGrades(b[sortBy]) - getAverageGrades(a[sortBy]);
 
+      case SortType.Married:
+      case SortType.Age:
+        return order === 'asc'
+          ? +a[sortBy] - +b[sortBy]
```
liliia-livandovskaya, line 46:
```suggestion
          ? Number(a[sortBy]) - Number(b[sortBy])
```
`Number` is more obvious way

liliia-livandovskaya (approved):
Great job!

PR #1221, opened 2023-04-03
function to sort students added

DenysKrysanov:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      ));
+      break;
+
+    default:
+      return copyOfStudents;
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return copyOfStudents;
```
Kavon-Skela, line 68: I think you can replace break instructions with return current sorted array and remove two additional returns
It also will allow you to write error handler in default case

Kavon-Skela (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+    case SortType.Age:
+      copyOfStudents.sort((a: Student, b: Student) => (isAsc
+        ? a[sortBy] - b[sortBy]
+        : b[sortBy] - a[sortBy]
+      ));
+      break;
+
+    case SortType.Married:
+      copyOfStudents.sort((a: Student, b: Student) => (isAsc
+        ? +a[sortBy] - +b[sortBy]
```
liliia-livandovskaya, line 52:
```suggestion
        ? Number(a[sortBy]) - Number(b[sortBy])
```
`Number()` is a more obvious way

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyOfStudents = [...students];
+  const isAsc: boolean = order === 'asc';
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      copyOfStudents.sort((a: Student, b: Student) => (isAsc
```
liliia-livandovskaya, line 37: `a, b`  bad variable names. You can use `firstStudent`, `secondStudent`

src/sortStudents.ts
liliia-livandovskaya: Agree

liliia-livandovskaya (changes requested):
Great job! 🔥 
Let's work on a fix

PR #1225, opened 2023-04-04
Solution

AnyaLyaLya:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
```
Evgenb000, line 16: I'm not sure exactly, but I think it should be like this
```suggestion
  Name,
  Surname,
  Age,
  Married,
  AverageGrade,
}

**Review by Evgenb000 — APPROVED** [review]
Good job!

File: src/sortStudents.ts (hunk 1 above)
Comment (liliia-livandovskaya) [reply src/sortStudents.ts:16]: no, its ok

File: src/sortStudents.ts [hunk 2]
```diff
@@ -1,16 +1,58 @@
...
+      case SortType.Married:
+        return order === 'desc'
+          ? secondStudent[sortBy].toString()
+            .localeCompare(firstStudent[sortBy].toString())
+          : firstStudent[sortBy].toString()
+            .localeCompare(secondStudent[sortBy].toString());
+
+      case SortType.Age:
+        return order === 'desc'
+          ? +secondStudent[sortBy] - +firstStudent[sortBy]
```
liliia-livandovskaya, line 46:
```suggestion
          ? Number(secondStudent[sortBy]) - Number(firstStudent[sortBy])
```
Number() is more obvious way

liliia-livandovskaya (approved):
Great job! 👍

PR #1227, opened 2023-04-04
solution

makandriival:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+      return Number(stud1Avg) - Number(stud2Avg);
+    }
+
+    if (order === 'desc') {
+      return Number(stud2[sortBy]) - Number(stud1[sortBy]);
+    }
+
+    return Number(stud1[sortBy]) - Number(stud2[sortBy]);
+  });
```
VenomMB, line 59: [CODE STYLE] - If you have a predetermined amount of options it is better to use a switch/case structure instead of multiple if/else blocks

VenomMB (changes requested):
Good job

liliia-livandovskaya (approved):

liliia-livandovskaya (approved):
🔥 🔥 🔥

PR #1230, opened 2023-04-04
solution

iHateAntichrist:


M1k1ta (approved):
Great!!!

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+const getAverage = (grades: number[]): number => {
```
liliia-livandovskaya, line 20:
```suggestion
const getAverageGrade = (grades: number[]): number => {
```

liliia-livandovskaya (changes requested):
Good job, but pay attention to the variable names. `value1, value2` are bad names without any context

liliia-livandovskaya (approved):

PR #1231, opened 2023-04-04
solution

glots-a:


Olekk17 (approved):

liliia-livandovskaya (changes requested):
Rewrite the code using witch/case. Checklist p.1

PR #1236, opened 2023-04-04
solution

Olekk17:


src/sortStudents.ts
```diff
@@ -1,16 +1,134 @@
...
+            student2: Student,
+          ) => student1.name.localeCompare(student2.name));
+          break;
+
+        case (SortType.Surname):
+          newStudentsArray.sort((
+            student1: Student,
+            student2: Student,
+          ) => student1.surname.localeCompare(student2.surname));
+          break;
```
kukla1989, line 42: combine them
3. [CODE STYLE] - DRY. Combine logic for sorting by fields with same data types.

kukla1989 (changes requested):
it would be better to use switch nested in if instead of switch nested in switch.
also you repeat to much code ( 29-77 lines and 80-130 lines) but honestly i dont know how to do this in you code

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+        return Number(student2[sortBy]) - Number(student1[sortBy]);
+      });
+      break;
+
+    case (SortType.AverageGrade):
+      newStudentsArray.sort((student1: Student, student2: Student) => {
+        const student1AverageGrade: number = student1.grades.reduce((
+          acc: number,
+          val: number,
+        ) => acc + val, 0) / student1.grades.length;
```
liliia-livandovskaya, line 61: would be better to create a separate function to calculate the `student average grade`

liliia-livandovskaya (approved):
Great job!

PR #1237, opened 2023-04-04
Solution

yuliia-anisimova:


Rage1502 (approved):
Good solution!!

liliia-livandovskaya (approved):
Great job! It would be better to avoid `a,b` names for the variables

PR #1242, opened 2023-04-04
Solution

danylo-dobrovolskyi:


StasLukianchuk (approved):
GJ

liliia-livandovskaya (approved):
Great job! It would be better to avoid `a,b, acc` names for the variables

PR #1246, opened 2023-04-04
Solution

Zad0rnik:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+    const avgB = b.grades.reduce((s, gr) => s + gr, 0) / b.grades.length;
+
+    switch (sortBy) {
+      case SortType.Name:
+        result = a.name.localeCompare(b.name);
+        break;
+
+      case SortType.Surname:
+        result = a.surname.localeCompare(b.surname);
+        break;
```
MylaCrystal, line 36: p. 3 of checklist: it can be shortened

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+
+      case SortType.Age:
+        result = a.age - b.age;
+        break;
+
+      case SortType.Married:
+        result = Number(a.married) - Number(b.married);
+        break;
```
MylaCrystal, line 44: and this one as well

MylaCrystal (changes requested):
Really adore this solution!

but I believe that you can fix these points according to the checklist a little bit more

src/sortStudents.ts
liliia-livandovskaya, line 36:
```suggestion
    switch (sortBy) {
      case SortType.Name:
      case SortType.Surname:
        result = a[sortBy].localeCompare(b[sortBy]);
        
        break;
```

src/sortStudents.ts
liliia-livandovskaya: agree

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments and the variable's names. `a,b, s, gr` are not the best choice.

src/sortStudents.ts
Zad0rnik: Снимок экрана 2023-04-05 в 11 40 49

liliia-livandovskaya (approved):

PR #1247, opened 2023-04-04
solution

StasLukianchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+): Student[] {
+  const studentsList = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      studentsList.sort((a: Student, b: Student) => (
+        order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy])
```
Stambuliak, line 42: do not use variables like 'a' or 'b'.  Maybe it is better to use student1 or firstStudent or prevStudent

Stambuliak (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
 export enum SortType {
-  // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
```
liliia-livandovskaya, line 18:
```suggestion
```

liliia-livandovskaya (approved):

PR #1257, opened 2023-04-05
Solution

Zibi95:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students:Student[],
+  sortBy:SortType,
+  order:SortOrder,
+):Student[] {
+  const sortedStudents = [...students];
+  let callback: (studentA:Student, studentB:Student) => number;
```
DorotaLeniecDev, line 27: I would suggest to name this function in more descriptive way so the code is more readable :)
Try to find better name

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      callback = (studentA, studentB):number => {
```
DorotaLeniecDev, line 32: I think that there is no need to define return type again
```suggestion
      callback = (studentA, studentB) => {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+  order:SortOrder,
+):Student[] {
+  const sortedStudents = [...students];
+  let callback: (studentA:Student, studentB:Student) => number;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
```
DorotaLeniecDev, line 31: I like this smart sollution

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      };
+      break;
+
+    default:
+      callback = (studentA, studentB):number => {
+        const studentASum
+          = studentA.grades.reduce((acc, curr) => acc + curr, 0);
+        const studentBSum
+          = studentB.grades.reduce((acc, curr) => acc + curr, 0);
+
```
DorotaLeniecDev, line 57: For readibility you could extract a function which will calculate average for each student grades

DorotaLeniecDev (changes requested):
Very good job! Check my few comments

src/sortStudents.ts
Zibi95: I wanted to avoid that too, but lint wasn't cooperating 
image. I guess I will disable that for entire file.

DorotaLeniecDev (approved):
Good job!

PR #1260, opened 2023-04-05
sort students solution

AP-Solution:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  sortedStudents.sort((a, b) => {
+    let comparison = 0;
+
+    switch (sortBy) {
+      case SortType.Name:
+        comparison = a.name.localeCompare(b.name);
```
AlexRog01, line 31: If you assign a value to each property in the enam, then you can use them as a key to the properties of the object.
`return student1[sortBy].localeCompare(student2[sortBy]);`
Then you don't need two cases for a first and last name

AlexRog01 (approved):
Good. But if you transfer the switch case from sorting, then you don’t have to check at each iteration which data needs to be compared. You can then select a property for comparison once, and sort by it

src/sortStudents.ts
SerhiiKirik: agree

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+        break;
+      case SortType.Age:
+        comparison = a.age - b.age;
+        break;
+      case SortType.Married:
+        comparison = Number(a.married) - Number(b.married);
+        break;
+      case SortType.AverageGrade: {
+        const averageA = a.grades
+          .reduce((sum, grade) => sum + grade, 0) / a.grades.length;
```
SerhiiKirik, line 44: DRY
```suggestion
        const averageA = compareAverage(a, b)
        const averageB = compareAverage(b, a)
```

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  sortedStudents.sort((a, b) => {
```
SerhiiKirik, line 26:
```suggestion
  sortedStudents.sort((firstStudent, secondStudent) => {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  function compareAverage(a: number[]): number {
+    return a
+      .reduce((sum: number, grade: number) => sum + grade, 0) / a.length;
+  }
```
jstmpelowycz, line 29:
```suggestion
  function calculateAverage(nums: number[]): number {
    const sum = nums.reduce((accum, num) => accum + num, 0);

    return sum / nums.length;
  }
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+  sortedStudents.sort((firstStudent, secondStudent) => {
+    let comparison = 0;
```
jstmpelowycz, line 32: do not make a value to reset, return in case immediately

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      case SortType.Surname:
+        comparison = firstStudent.surname.localeCompare(secondStudent.surname);
+        break;
+      case SortType.Age:
+        comparison = firstStudent.age - secondStudent.age;
+        break;
+      case SortType.Married:
+        comparison = Number(firstStudent.married)
+          - Number(secondStudent.married);
+        break;
```
jstmpelowycz, line 47: combine cases, as you can see, you sort strings and numbers, but repeat yourself

jstmpelowycz (changes requested):

PR #1266, opened 2023-04-05
fix

asmd564:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+      );
+
+    case SortType.Married:
+
+      return sortedStudents.sort(
+        (a: Student, b: Student) => {
+          return (Number(a.married) - Number(b.married))
+          * sortOrder;
+        },
+      );
```
jstmpelowycz, line 60:
```suggestion
    case SortType.Age:
      return sortedStudents.sort(
        (a: Student, b: Student) => {
          return (a.age - b.age) * sortOrder;
        },
      );

    case SortType.Married:
      return sortedStudents.sort(
        (a: Student, b: Student) => {
          return (Number(a.married) - Number(b.married))
          * sortOrder;
        },
      );
```

jstmpelowycz: I believe, it's possible to combine this cases into one, for example
```js
case SortType.Age:
case SortType.Married:
  return sortedStudents.sort(
    (studentA: Student, studentB: Student) => (
      (Number(studentA[sortBy]) - Number(studentB[sortBy])) * sortOrder
    )
  );
```
and the same way for strings, so you just sort string and numbers

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+
+    case SortType.AverageGrade:
+
+      return sortedStudents.sort(
+        (a: Student, b: Student) => {
+          const avrA = a.grades.reduce((sum, grade) => sum + grade)
+          / a.grades.length;
```
jstmpelowycz, line 67: would be perfect to use a dedicated function to calculate average number

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+):Student[] {
+  const sortedStudents: Student[] = [...students];
+  const sortOrder = order === 'asc' ? 1 : -1;
+
+  switch (sortBy) {
+    case SortType.Name:
+      sortedStudents.sort(
```
jstmpelowycz, line 30: you should return this value

jstmpelowycz (changes requested):

PR #1279, opened 2023-04-18
solution

Nikolaymelni:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
-  // write your function
+  return result;
+};
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsArrayCopy: Student[] = [...students];
```
qoa11a, line 34: No need to specify the type here. TypeScript understands it itself.

qoa11a (changes requested):
Not bad! 👍

polosanya (approved):
Well done!

PR #1281, opened 2023-04-18
Solution

Vaskovics:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
-  // write your function
+  return everage;
+}
+
+export function sortStudents(
+  students: Student,
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsArr: Student[] = [...students];
```
loralevitska, line 32:
```suggestion
  const studentsCopy: Student[] = [...students];
```

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    case SortType.Age:
+      studentsArr.sort((a, b) => (isAsc
+        ? a[sortBy] - b[sortBy]
+        : b[sortBy] - a[sortBy]));
+      break;
+
+    case SortType.AverageGrade:
+      studentsArr.sort((a: Student, b: Student): number => (isAsc
+        ? getAverageGrade(a) - getAverageGrade(b)
+        : getAverageGrade(b) - getAverageGrade(a)));
```
loralevitska, line 53: Combine this methods with helper function `sortByNumber`, also here use order

loralevitska (changes requested):
A little more and you will be on target)

src/sortStudents.ts
Vaskovics: that will be quite complicated function where I need to use anther helper function. Sorry I could't finish this request.

OleziO (approved):

PR #1286, opened 2023-04-19
solution

mateusz-michalowski:


natalia-klonowska (changes requested):
you have descending order only for averageGrade

DanielSorokowski (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
 }
 
 export enum SortType {
   // describe SortType enum
+  Name,
```
d-wasilewski, line 14:
```suggestion
  Name = 'name',
```
etc

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+        case 'asc':
+          copiedStudents.sort((a, b) => a.surname.localeCompare(b.surname));
+          break;
+        case 'desc':
+          copiedStudents.sort((a, b) => b.surname.localeCompare(a.surname));
+          break;
+        default:
+          throw new Error(`Type 'asc' or 'desc' not ${order}`);
+      }
+      break;
```
d-wasilewski, line 56: connect similar cases like these 
```
    case SortType.Name:
    case SortType.Surname:
```
instead of a.surname, you can use a[sortBy]
also, change the naming from a, b to something more specific

d-wasilewski (changes requested):

Amsko6 (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,99 @@
...
+        case 'desc':
+          copiedStudents.sort(
+            (prevStudent, currStudent) => currStudent[sortBy]
+              .localeCompare(prevStudent[sortBy]),
+          );
+          break;
+        default:
+          throw new Error(`Type 'asc' or 'desc' not ${order}`);
+      }
+      break;
```
d-wasilewski, line 51: you don't have to nest switch cases like this, use ternary operator with only 2 options, also one default statement is enough at the end
`'asc' ? case a : case b`

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1290, opened 2023-04-19
first solution

Ozwiena:


src/sortStudents.ts
```diff
@@ -1,16 +1,89 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
+  const studentsCopy = [...students];
+  let result: Student[] = [];
```
d-wasilewski, line 24: checklist 4

src/sortStudents.ts
```diff
@@ -1,16 +1,89 @@
...
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      result = (order === 'asc')
+        ? studentsCopy.sort(
+          (a: Student, b: Student) => a.name.localeCompare(b.name),
+        )
+        : studentsCopy.sort(
+          (a: Student, b: Student) => b.name.localeCompare(a.name),
+        );
```
d-wasilewski, line 34: connect similar cases like these 
instead of a.surname, you can use a[sortBy]
also, change the naming from a, b to something more specific

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1296, opened 2023-04-20
solution

KielpinskiJ:


Ozwiena (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,116 @@
...
+  grades: number[],
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
```
d-wasilewski, line 15:
```suggestion
  Name = 'name',
  Surname = 'surname',
  Age = 'age',
  Married = 'married',
  AverageGrade = 'averageGrade',
```

src/sortStudents.ts
```diff
@@ -1,16 +1,116 @@
...
-  // write your function
+  switch (order) {
+    case 'asc': {
+      switch (sortBy) {
+        case SortType.Name: {
+          studentsCopy.sort((a, b) => a.name.localeCompare(b.name));
+          break;
+        }
+
+        case SortType.Surname: {
```
d-wasilewski, line 35: connect similar cases (checklist 3 - DRY)
```
case SortType.Name:
case SortType.Surname:
```
as a key you can use `sortBy` property

d-wasilewski: also change the naming from `a, b to` something more specific

d-wasilewski (changes requested):

PR #1297, opened 2023-04-21
solution

SuaBy514:


src/sortStudents.ts
```diff
@@ -1,16 +1,119 @@
...
+        }
+
+        if (avgB > avgA && order === 'desc') {
+          return 1;
+        }
+
+        return 0;
+      });
+      break;
+    }
```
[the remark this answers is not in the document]
SuaBy514: Reduce is anything but simple hahaha
I will look into it

src/sortStudents.ts
```diff
@@ -1,16 +1,102 @@
...
+
+    case SortType.Age:
+      if (order === 'asc') {
+        clonedStudents.sort((student, student2) => student.age - student2.age);
+        break;
+      }
+
+      clonedStudents.sort((student, student2) => student2.age - student.age);
+      break;
+
```
d-wasilewski, line 57: these 2 can be connected too

src/sortStudents.ts
```diff
@@ -1,16 +1,102 @@
...
+    case SortType.Name:
+    case SortType.Surname:
+      if (order === 'asc') {
+        clonedStudents.sort(
+          (student, studen2) => student[sortBy].localeCompare(studen2[sortBy]),
+        );
+        /* Please excuse the variable name studen2, had issues with
+        max length and typescript seems to not like line breaks in
+        arrow functions (the linter kept yelling at me) */
+        break;
```
d-wasilewski, line 41: Have you tried something like this?
```suggestion
        clonedStudents.sort(
          (student, secondStudent) => student[sortBy]
            .localeCompare(studen2[sortBy]),
        );
        break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,102 @@
...
+        clonedStudents.sort(
+          (a, b) => String(a.married).localeCompare(String(b.married)),
+        );
+        // Here the line is already too long to change variable names
+        break;
+      }
+
+      clonedStudents.sort(
+        (a, b) => String(b.married).localeCompare(String(a.married)),
+      );
```
d-wasilewski, line 69: maybe use ternary operator for this

d-wasilewski (changes requested):

src/sortStudents.ts
SuaBy514: Looks wild to me, I will give it a try

src/sortStudents.ts
SuaBy514: I'm afraid ternary operators are the one thing in JavaScript that's black magic to me, every time I try to use them, they just do not function the way I expect them to

src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+      if (order === 'asc') {
+        clonedStudents.sort(
+          (student, student2) => +student[sortBy] - +student2[sortBy],
+        );
+        break;
+      }
+
+      clonedStudents.sort(
+        (student, student2) => +student2[sortBy] - +student[sortBy],
+      );
```
d-wasilewski, line 59: It probably would look something like this with ternary operators. Try it out and if you have any questions about how it works, you can come to the QnA. It might be a black magic, but they are very usefull
```suggestion
      return order === 'asc'
        ?  clonedStudents.sort(
          (student, student2) => +student[sortBy] - +student2[sortBy],
        );
        : clonedStudents.sort(
        (student, student2) => +student2[sortBy] - +student[sortBy],
      );
```

src/sortStudents.ts
```diff
@@ -1,16 +1,93 @@
...
+      clonedStudents.sort((stdnt, stdnt2) => {
+        let avg1: number = stdnt.grades.reduce((prv, cur) => prv + cur, 0);
+        let avg2: number = stdnt2.grades.reduce((prv, cur) => prv + cur, 0);
+
+        avg1 /= stdnt.grades.length;
+        avg2 /= stdnt2.grades.length;
+
+        return avg2 - avg1;
+      });
+      break;
```
d-wasilewski, line 86: unify the code

d-wasilewski (changes requested):

src/sortStudents.ts
SuaBy514: Doing this causes the sort to not work

SuaBy514: ..Nevermind, I was looking at the wrong line

d-wasilewski (approved):

PR #1299, opened 2023-04-21
solution

popeye88:
Please, suggest me right code style in my cases. TY


OleziO (approved):

popeye88:
Could you please advice me if my code style is correct or give some suggestions on it

PR #1302, opened 2023-04-22
commit message

EvgenyNikitinEvgeny:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  Married,
+  AverageGrade,
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = {
+  asc : string,
+  desc: string,
+};
```
ivbutsykin, line 23: I suggest rewrite in this way:
```suggestion
export type SortOrder  = 'asc' | 'desc';
```

ivbutsykin (changes requested):

ivbutsykin (approved):
Nice, good job

PR #1303, opened 2023-04-22
FIX

marshalek37:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(students: Student[],
+  sortBy: SortType,
+  order: SortOrder): Student[] {
+  function compareAvgGrades(firstStudent: Student,
```
d-wasilewski, line 22: this function can be moved outside of the `sortStudents` function body

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+    = firstStudent.grades.reduce((acc: number, val: number) => acc + val)
+    / firstStudent.grades.length;
+    const avgGradeB
+    = secondStudent.grades.reduce((acc: number, val: number) => acc + val)
+    / secondStudent.grades.length;
+
+    return avgGradeA - avgGradeB;
+  }
+
+  const sortedStudents = [...students];
```
d-wasilewski, line 34: they are not sorted here yet, they are just copied

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        comparison = firstStudent[sortBy].localeCompare(secondStudent[sortBy]);
+        break;
+      case SortType.Age:
+        comparison = firstStudent.age - secondStudent.age;
+        break;
+      case SortType.Married:
+        if (firstStudent.married === secondStudent.married) {
+          comparison = 0;
```
d-wasilewski, line 49: it's 0 by default, so you don't need to overwrite it here, try to simplify this statement to use only the second expression (` comparison = firstStudent.married ? 1 : -1;
`)

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1308, opened 2023-04-25
solve

ctgdprzyb:


src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const resultArray = [...students];
+
+  const average = (arr: number[]): number => {
+    return arr.reduce((p: number, c: number) => p + c, 0) / arr.length;
+  };
 
+  switch (sortBy) {
+    case 'name':
```
d-wasilewski, line 43: Use `SortType.name` instead of 'name'

src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+      if (order === 'desc') {
+        resultArray.sort(
+          (a: Student, b: Student): number => b.name.localeCompare(a.name),
```
d-wasilewski, line 46: these types are probably infered and you don't need to write them
```suggestion
          (a, b): number => b.name.localeCompare(a.name),
```

src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+        );
+      } else {
+        resultArray.sort(
+          (a: Student, b: Student): number => a.name.localeCompare(b.name),
+        );
+      }
+      break;
+    case 'surname':
```
d-wasilewski, line 54: join similar cases (like name and surname)

src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+          : number => average(b.grades) - average(a.grades),
+        );
+      } else {
+        resultArray.sort(
+          (a: Student, b: Student)
+          : number => average(a.grades) - average(b.grades),
+        );
+      }
+      break;
+    default:
```
d-wasilewski, line 100: throw an error here

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1329, opened 2023-05-12
initial commit

44shaggy:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  type Compare = string | number | boolean;
+
+  function compare(x: Student): Compare {
+    switch (sortBy) {
+      case 'name':
+      case 'surname':
```
anstsot, line 34: use `SortType`
```suggestion
      case SortType.Name:
      case SortType.Surname:
```

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      case 'married':
+        return x[sortBy];
+      case 'averageGrade':
+      default:
+        return x.grades.reduce((a, b) => a + b, 0) / x.grades.length;
+    }
+  }
+
+  return students.slice().sort((a, b) => {
+    switch (true) {
```
anstsot, line 46: it's not the best solution to use `switch (true)`. try to use `if`

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      case 'age':
+      case 'married':
+        return x[sortBy];
+      case 'averageGrade':
+      default:
+        return x.grades.reduce((a, b) => a + b, 0) / x.grades.length;
+    }
+  }
+
+  return students.slice().sort((a, b) => {
```
anstsot, line 45: why do you use `slice`?
```suggestion
  return [...students].sort((a, b) => {
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/sortStudents.ts
44shaggy: but slice() is one way to copy array taught in the course video ;P

src/sortStudent.test.js
```diff
@@ -0,0 +1,199 @@
+"use strict";
+/* eslint-disable max-len */
```
Anastasiia-Svintsova, line 2: Why did you change this file? Remove all changes here

src/sortStudents.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+var __spreadArray = (this && this.__spreadArray) || function (to, from, pack) {
```
Anastasiia-Svintsova, line 2: Here should be only sortStudents.ts file. Remove this file as well

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  function setOrder(): number {
```
Anastasiia-Svintsova, line 25: Get used to using arrow functions. Arrow functions have a more concise syntax than traditional function declarations, which makes them easier to read and write.

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+    return order === 'asc' ? 1 : -1;
+  }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  type Compare = string | number | boolean;
+
+  function compare(x: Student): Compare {
```
Anastasiia-Svintsova, line 31: Don't declare types inside the function. And you don't really need a separate type here. Improve variable names 
```suggestion
  function compare(student: Student): string | number {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+        return x.grades.reduce((a, b) => a + b, 0) / x.grades.length;
+    }
+  }
+
+  return [...students].sort((a, b) => {
+    if (compare(a) > compare(b)) {
+      return setOrder();
+    }
+
+    if (compare(a) < compare(b)) {
```
Anastasiia-Svintsova, line 50: You calculate `compare(a)` and `compare(b)` twice. Move these values to variables, to avoid extra function calls.

Anastasiia-Svintsova (changes requested):

agniya-i (approved):

PR #1330, opened 2023-05-12
solution

Nikoramo:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+        return updateStudent.sort((a, b) => b.age - a.age);
+      case SortType.Married:
+        return updateStudent.sort((a, b) => Number(b.married)
+    - Number(a.married));
+      case SortType.AverageGrade:
+        return updateStudent.sort((a, b) => b.averageGrade
+        - a.averageGrade);
+      default:
+        return [];
+    }
```
sTorba24, line 68: i'm sure you can minimize it, use ternary operator to change order (a,b) in expression

Also you can combine some of cases because expression inside is the same

```
    case SortType.Name:
    case SortType.Surname:
      studentsArray.sort((a, b) => (order === 'asc'
        : b[sortBy].localeCompare(a[sortBy])
      ));
      break;
 ```     
 
 it should be something like this

sTorba24 (changes requested):
Good work! Lets improve it

PR #1331, opened 2023-05-12
Initial commit

RomanHrytsiuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+  if (sortBy === SortType.AverageGrade && order === 'desc') {
+    return copy.sort(
+      (a: Student, b: Student) => getAverageGrade(b.grades)
+        - getAverageGrade(a.grades),
+    );
+  }
+
+  if (sortBy === SortType.AverageGrade && order === 'asc') {
+    return copy.sort(
+      (a: Student, b: Student) => getAverageGrade(a.grades)
```
SanyaBratashchuk, line 65: use switch case

SanyaBratashchuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+      break;
+    case SortType.Married:
+      if (order === 'desc') {
+        return copy.sort(
+          (a: Student, b: Student) => Number(b.married) - Number(a.married),
+        );
+      }
+      break;
+    case SortType.AverageGrade:
+      if (order === 'desc') {
```
TarasHoliuk, line 62: you may use ternary operator instead of `if` block to make it look more laconic

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+        );
+      }
+      break;
+    case SortType.Surname:
+      if (order === 'asc') {
+        return copy.sort(
+          (a: Student, b: Student) => a.surname.localeCompare(b.surname),
+        );
+      }
+      break;
```
TarasHoliuk, line 46: you may join a few cases in this way:
```suggestion
    case SortType.Name:
    case SortType.Surname:
      if (order === 'asc') {
        return copy.sort(
          (a: Student, b: Student) => a[sortBy].localeCompare(b[sortBy]),
        );
      }
      break;
```
the same with Age and Married (use Married code, with transforming to `number`)
the only thing you have to do to make it possible is to add values in the enum which match with keys in `student` object, example:
```
enum Example {
  Name = 'name'
}
```

TarasHoliuk (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const copy: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? copy.sort(
+          (a: Student, b: Student) => a[sortBy].localeCompare(b[sortBy]),
+        )
+        : copy;
```
tetiana-ivanchenko, line 40: if order is 'desc' you should sort too, but in different order

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'desc'
+        ? copy.sort(
+          (a: Student, b: Student) => Number(b[sortBy]) - Number(a[sortBy]),
+        )
+        : copy;
```
tetiana-ivanchenko, line 48: the same

tetiana-ivanchenko (changes requested):

anstsot (approved):
Good job, approved

PR #1334, opened 2023-05-15
TS... pain

victor-buldenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,90 @@
...
+): Student[] {
+  function copyObj(obj: Student): Student {
+    return Object.assign(obj);
+  }
+
+  const copy = students.reduce((res: Student[], el) => {
+    res.push(copyObj(el));
+
+    return res;
+  }, [] as Student[]);
```
SanyaBratashchuk, line 32: why do you need to make a copy of each student?

src/sortStudents.ts
```diff
@@ -1,16 +1,90 @@
...
+  }
+
+  if (sortBy === SortType.Surname) {
+    copy.sort((a, b) => {
+      if (order === 'asc') {
+        return a.surname.localeCompare(b.surname);
+      }
+
+      return b.surname.localeCompare(a.surname);
+    });
```
SanyaBratashchuk, line 51: use switch case

SanyaBratashchuk (changes requested):

src/sortStudents.ts
victor-buldenko: because i thought, function should return full copy array & contents.

DarkMistyRoom (approved):

PR #1336, opened 2023-05-16
add sortStudents solution

DidenkoDenis306:


src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  return grades.reduce((acc, value) => (
+    acc + value)) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students];
```
dinaminahi, line 30: You can omit `Student[]` typing since typescript can handle it, because you already give the type for `students`

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
+    return studentsCopy.sort((student1: Student, student2: Student) => (
```
dinaminahi, line 33: Same here, you don't really have to specify the type for `student1` and `student2`, typescript will know it by itself, because you are sorting `studentsCopy` which is `Student` array.

In practice you should specify the type when:

- When you declare a variable and assign it a value later.
- When you want a variable that can’t be inferred.
- When a function returns the `any` type and you need to clarify the value.

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
```
dinaminahi, line 20: Same here about  type that function return
https://www.typescripttutorial.net/typescript-tutorial/typescript-type-inference/

dinaminahi (changes requested):

src/sortStudents.ts
DidenkoDenis306: I know this, but linter forces me to write the return type of the function, otherwise it writes an error

dinaminahi: Okay, my fault, sorry

dinaminahi (approved):
Well done 🥇

PR #1337, opened 2023-05-17
Develop

generatedman:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
```
dinaminahi, line 24: So here is a great example, you didn't specify the type for `sortedStudents`  and typescript know that it has type `Student[]` because it is a copy of array with that type.

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+
+  function calculateAverage(numbers: number[]): number {
+    return numbers.reduce((acc, curr) => acc + curr, 0) / numbers.length;
+  }
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return sortedStudents.sort((a: Student, b: Student) => {
```
dinaminahi, line 33: And here for example typescript will also know the type of `a` and `b`, so no need to specify them. By the way, would be better to create more descriptive names, instead of a and b

dinaminahi (approved):
Great Job! 

There is `type inference` in Typescript, it is when Typescript can handle the type by itself
In practice, you should always use the type inference as much as possible. And specify the type in the following cases:

- When you declare a variable and assign it a value later.
- When you want a variable that can’t be inferred.
- When a function returns the any type and you need to clarify the value.

https://www.typescripttutorial.net/typescript-tutorial/typescript-type-inference/

PR #1341, opened 2023-05-21
add task solution

maksym-nemera:


src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents: Student[] = [...students];
+
+  return sortedStudents.sort((a, b) => {
+    switch (sortBy) {
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? a.grades
```
33Orange, line 30: Extract the average grade calculation into a separate function to improve code readability and maintainability.

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+          : Number(b[sortBy]) - Number(a[sortBy]);
+
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
 
+      default:
+        break;
```
33Orange, line 50: Remove the unnecessary 'break' statement in the default case.
It's better to add error handling if your sorting is in the default case.

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+    }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    return 0;
```
33Orange, line 53: Remove the unnecessary 'return 0;' line as the function already returns the sorted array.

33Orange (changes requested):
Almost done!

A few moments to fix.

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents: Student[] = [...students];
+
+  return sortedStudents.sort((a, b) => {
```
dinaminahi, line 31: It is better to create more descriptive names - a and b tell nothing 
Something like student1 and student2 or firstStudent and secondStudent will be better

dinaminahi (approved):
Well done 💪

PR #1342, opened 2023-05-22
solution

oksanatytanych:


src/sortStudents.ts
```diff
@@ -1,16 +1,89 @@
...
+      case SortType.AverageGrade:
+        sortedStudents.sort(
+          (a: Student, b: Student) => sortArray(a.grades) - sortArray(b.grades),
+        );
+        break;
+      default:
+        throw new Error('Error...');
+    }
+  } else {
+    switch (sortBy) {
```
Viktor-Kost, line 59: 1) instead of using if/else you can apply ternary inside each case.
2) no needs to add types for `a` and `b` as they are inferred 
3) you can combine similar groups of sorting
Example: 
```
      case SortType.Name:
      case SortType.Surname:
        return order === 'asc' 
        ? sortedStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
        : sortedStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```

Viktor-Kost (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
 
+  function sortArray(array: number[]): number {
```
etojeDenys, line 26: The function name 'sortArray' is misleading since it calculates the average, not sorting. Consider renaming it to 'calculateAverage'.

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #1345, opened 2023-05-24
Solved

silcki:


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+
+  switch (sortBy) {
+    default:
+    case SortType.Age:
+      if (order === 'desc') {
+        sortFunction = (a, b): number => {
+          return b.age - a.age;
+        };
+      }
+      break;
```
IvanFesenko, line 42: How can I sord in `asc` ordering?

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+      if (order === 'asc') {
+        sortFunction = (a, b): number => {
+          return +a.married - +b.married;
+        };
+      } else {
+        sortFunction = (a, b): number => {
+          return +b.married - +a.married;
+        };
+      }
+      break;
```
IvanFesenko, line 65: create sortFunction without reasign. Move logic inside function. Instead of usign if else statement use ternary operator

```const sortFunction = (a: Student, b: Student): number => {
    switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
        return asc
```

IvanFesenko (changes requested):
Lets improve your code. Check my comments

mykhalenych (approved):

PR #1352, opened 2023-05-28
add task solution

KatyaZaremska:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(student: Student): number {
+  const averageGrade: number = student.grades.reduce(
+    (a: number, b: number) => a + b, 0,
+  )
+    / student.grades.length;
```
33Orange, line 24: Why take the entire student object as an argument if we only want the grades array?
What are the names for the reduce method arguments?

```suggestion
function getAverageGrade(grades: number[]): number {
  return grades.reduce((acc, grade) => (
    acc + grade), 0) / grades.length;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedObject: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      sortedObject.sort((a: Student, b: Student): number => {
```
33Orange, line 39: Use more descriptive names for arguments in sort callback.

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        if (order === 'desc') {
+          return b[sortBy].localeCompare(a[sortBy]);
+        }
+
+        return a[sortBy].localeCompare(b[sortBy]);
```
33Orange, line 44: Good case to use ternary operator

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      break;
+
+    case SortType.Married:
+    case SortType.Age:
+      sortedObject.sort((a: Student, b: Student): number => {
+        if (order === 'desc') {
+          return +b[sortBy] - +a[sortBy];
+        }
+
+        return +a[sortBy] - +b[sortBy];
```
33Orange, line 55: Same as in previous case

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+      break;
+
+    case SortType.AverageGrade:
+      sortedObject.sort((a: Student, b: Student): number => {
+        if (order === 'desc') {
+          return getAverageGrade(b) - getAverageGrade(a);
+        }
+
+        return getAverageGrade(a) - getAverageGrade(b);
```
33Orange, line 65: Same as in previous cases

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+      break;
+
+    default:
```
33Orange, line 69: Use `default` case for error handling.
In what cases can you get an error here?

33Orange (changes requested):
Good job.

A few moments to fix

naziamoff (changes requested):
great job! 
Now let's try to integrate our sorting by grades with other sorting ways inside the switch statement

naziamoff (approved):
🔥

PR #1358, opened 2023-05-30
Solution

amashutikov:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+      (order === 'asc')
+        ? studentA.name.localeCompare(studentB.name)
+        : studentB.name.localeCompare(studentA.name)
+    ));
+  } else if (sortBy === SortType.Surname) {
+    copyOfStudentsArray.sort((studentA, studentB) => (
+      (order === 'asc')
+        ? studentA.surname.localeCompare(studentB.surname)
+        : studentB.surname.localeCompare(studentA.surname)
+    ));
```
33Orange, line 44: You can combine this 2 cases

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+      (order === 'asc')
+        ? studentA.age - studentB.age
+        : studentB.age - studentA.age
+    ));
+  } else if (sortBy === SortType.Married) {
+    copyOfStudentsArray.sort((studentA, studentB) => (
+      (order === 'asc')
+        ? Number(studentA.married) - Number(studentB.married)
+        : Number(studentB.married) - Number(studentA.married)
+    ));
```
33Orange, line 56: Same as in previous. You can combine this 2 cases

33Orange (changes requested):
Good job.

A few moments :
1. Let's use `switch` case instead if. Checklist p1.
And dont forget about error handling in `default` case 
[CODE STYLE] - If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks

2. You can compare some sort types. Checklist p3.
`[CODE STYLE] - DRY. Combine logic for sorting by fields with same data types.`

Everything else is fine

33Orange (approved):
Well done

PR #1359, opened 2023-05-30
solution

Yuliya1907:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+function getAverageGrade(student: Student): number {
+  const averageGrade: number = student.grades.reduce(
+    (a: number, b:number) => a + b, 0,
+  ) / student.grades.length;
+
+  return averageGrade;
+}
```
33Orange, line 29: Why take the entire student object as an argument if we only want the grades array?
What are the names for the reduce method arguments?

```suggestion
function getAverageGrade(grades: number[]): number {
  return grades.reduce((acc, value) => (
    a + b), 0) / grades.length;
}
```

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedObject: Student[] = [...students];
+
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      sortedObject.sort((a: Student, b: Student): number => {
```
33Orange, line 41: What are the names for the sort method arguments?
Use more descriptive names

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        if (order === 'desc') {
+          return b[sortBy].localeCompare(a[sortBy]);
+        }
+
+        return a[sortBy].localeCompare(b[sortBy]);
```
33Orange, line 46: Good case to use ternary operator. Let's change it

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      break;
+
+    case SortType.Married:
+    case SortType.Age:
+      sortedObject.sort((a: Student, b: Student): number => {
+        if (order === 'desc') {
+          return +b[sortBy] - +a[sortBy];
+        }
+
+        return +a[sortBy] - +b[sortBy];
```
33Orange, line 57: Same as in first sort method. Change arguments names, use ternary operator

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+      break;
+    case SortType.AverageGrade:
+      sortedObject.sort((a: Student, b: Student): number => {
+        if (order === 'desc') {
+          return getAverageGrade(b) - getAverageGrade(a);
+        }
+
+        return getAverageGrade(a) - getAverageGrade(b);
```
33Orange, line 66: Same as in previous cases

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      });
+      break;
+    default:
```
33Orange, line 69: Use `default` case for error handling.

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+
+  return averageGrade;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedObject: Student[] = [...students];
```
33Orange, line 36: Use more descriptive names.
Don't use object type names in their name

33Orange (changes requested):
Good job.

A few moments to fix

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+  return grades.reduce((acc, value) => (
+    acc + value), 0) / grades.length;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedObject = [...students];
```
33Orange, line 33: Use more descriptive names and don't use object types in their names
```suggestion
  const copyOfStudents = [...students];
```

33Orange (approved):
Well done !

PR #1360, opened 2023-05-30
Add task solution

IrynaBalandiukh:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  const studentsCopy = [...students];
+
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
+    return order === 'asc'
+      ? studentsCopy.sort(
+        (current, next) => current[sortBy].localeCompare(next[sortBy]),
+      )
+      : studentsCopy.sort(
+        (current, next) => next[sortBy].localeCompare(current[sortBy]),
+      );
```
33Orange, line 38: You can always return sorted array and change sorting inside callback when you have different conditions
```suggestion
    return studentsCopy.sort((current, next) => (
      sortType === 'asc'
        ? current[sortBy].localeCompare(next[sortBy])
        : next[sortBy].localeCompare(current[sortBy])
    ));
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  }
+
+  if (sortBy === SortType.Age || sortBy === SortType.Married) {
+    return order === 'asc'
+      ? studentsCopy.sort(
+        (current, next) => +current[sortBy] - +next[sortBy],
+      )
+      : studentsCopy.sort(
+        (current, next) => +next[sortBy] - +current[sortBy],
+      );
```
33Orange, line 48: Same as in previous case

33Orange (changes requested):
Good job.

A few moments to fix:
1. Let's use `switch` case instead `if`. Checklist p1. 
And don't forget about error handling in `default` case when use `switch`
`If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks`

2. Change logic in first 2 cases. Let's use only one sort method and change the conditions inside its callback

33Orange (approved):
Well done !

PR #1362, opened 2023-05-31
task solution

hekaq:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(grades: number[]): number {
```
33Orange, line 20:
```suggestion
function getAverageGrade(grades: number[]): number {
```

33Orange (changes requested):
Good job.

A few moments:
1. You can use 1 sort method and use condition inside.
2. Use `switch` case instead if. Checklist p1.
`[CODE STYLE] - If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks`

Everything else is fine

naziamoff (approved):
🔥

PR #1364, opened 2023-05-31
add solution

growingananas:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const newStudentsArr: Student[] = [...students];
```
naziamoff, line 25: we never add the data type to the name of our variable, at least on this level of abstraction. We're dealing with the students, and in this variable we store the copy of the students.

naziamoff (changes requested):
let's implement the whole sorting logic inside a callback of a single sort

naziamoff (approved):
looks good 👍 
But pls be more attentive next time, as you didn't fix all of my comments

growingananas:
> looks good 👍 But pls be more attentive next time, as you didn't fix all of my comments

Hey! Thank you)
Which comments do you mean in particular? 
I have fixed this 'let's implement the whole sorting logic inside a callback of a single sort' and this 'we never add the data type to the name of our variable, at least on this level of abstraction. We're dealing with the students, and in this variable we store the copy of the students.'  Have I missed anything? Looking forward to your responce))

naziamoff:
>'we never add the data type to the name of our variable, at least on this level of abstraction. We're dealing with the students, and in this variable we store the copy of the students.'

Yeah, this comment is actually about the naming. You still have it as a `newStudentsArr`, but this name does not give us any info. Why is it new? Are we going to create new students? Or add the existing ones?

So to make it maximally clear I would use `studentsCopy` - so there were no questions left regarding the purpose of this variable

growingananas:
> > 'we never add the data type to the name of our variable, at least on this level of abstraction. We're dealing with the students, and in this variable we store the copy of the students.'
> 
> Yeah, this comment is actually about the naming. You still have it as a `newStudentsArr`, but this name does not give us any info. Why is it new? Are we going to create new students? Or add the existing ones?
> 
> So to make it maximally clear I would use `studentsCopy` - so there were no questions left regarding the purpose of this variable

oh, I see now. You are right, I have not fixed this issue cause I didn't get your idea. I will do it now. Thank you a lot)

PR #1366, opened 2023-05-31
add solution

ivankhrsh:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name = 'NAME',
+  Surname = 'SURNAME',
+  Age = 'AGE',
+  Married = 'MARRIED',
+  AverageGrade = 'AVERAGE_GRADE',
```
naziamoff, line 15: WHY U YELLING AT ME 😱 

Jokes aside - not critical, but I don't see a reason for making it in upper case here

naziamoff (changes requested):
you repeat the same logic in different cases, e.g when you compare by name and surname. Also, let's use ternary and assign the callback by order. Also, let's write our callback inside the sort method - it will look more clear and concise

src/sortStudents.ts
ivankhrsh: I did it same as in checklist file:
image

naziamoff (approved):
awesome 👍

PR #1367, opened 2023-05-31
add functionality to sortStudents

cephy77:


naziamoff (changes requested):
let's have a single sort and implement the whole logic inside the callback

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
+  function sortInOrder(num: number): number {
+    return order === 'asc' ? num : -num;
+  }
+
+  function averageGrade(grades: number[]): number {
```
naziamoff, line 32: functions start with a verb - it must be clear we're dealing with the function by its name

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+
+      case SortType.AverageGrade:
+        return sortInOrder(averageGrade(a[sortBy]) - averageGrade(b[sortBy]));
+
+      default:
+        throw new Error('Something wrong, I can feel it');
+    }
+  };
 
+  studentsCopy.sort(sortingCallBack);
```
naziamoff, line 54: I would have it explicitly instead. `.sort((a, b) => {
...logic goes here
})`

Not critical suggestion though

naziamoff (approved):
good job 🔥

src/sortStudents.ts
dinaminahi, line 32: It is recommended to start function names with verb, for example `getAverageGrade`

dinaminahi, line 32: Let's move these helper functions outside `sortStudents`, to make it more readable

src/sortStudents.ts
dinaminahi, line 54: Let's return it here

src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+
+      case SortType.Age:
+      case SortType.Married:
+        return sortInOrder(Number(a[sortBy]) - Number(b[sortBy]));
+
+      case SortType.AverageGrade:
+        return sortInOrder(averageGrade(a[sortBy]) - averageGrade(b[sortBy]));
+
+      default:
+        throw new Error('Something wrong, I can feel it');
```
dinaminahi, line 50: Cool error 😅 But consider of making it more descriptive, include wrong sortBy here

dinaminahi (changes requested):
Good job 💪 
Few fixes needed to improve it

naziamoff (approved):

PR #1371, opened 2023-05-31
solution

dimoshen:


src/sortStudents.ts
```diff
@@ -1,16 +1,92 @@
...
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverageGrade(grades: number[]): number {
+  if (grades.length === 0) {
+    return 0;
+  }
```
naziamoff, line 25: what happens if you don't make this validation?

src/sortStudents.ts
```diff
@@ -1,16 +1,92 @@
...
+
+function sortByAverageGrade(a: Student, b: Student, order: SortOrder): number {
+  const avgGradesA = calculateAverageGrade(a.grades);
+  const avgGradesB = calculateAverageGrade(b.grades);
+
+  if (order === 'asc') {
+    return avgGradesA - avgGradesB;
+  }
+
+  return avgGradesB - avgGradesA;
```
naziamoff, line 40: ternary

naziamoff (changes requested):
let's have just a single `sort` and implement the whole logic inside the callback

src/sortStudents.ts
dimoshen: deleted

src/sortStudents.ts
dimoshen: fixed

naziamoff (changes requested):
image

PR #1372, opened 2023-05-31
Solution

annholovko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrades(grades: number[]): number {
```
naziamoff, line 20: function name always starts with a verb. It should have clear meaning

naziamoff (approved):
looks good 👍

PR #1373, opened 2023-05-31
made ts task sort students

vladyslav-lapshyn:


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+        }
+
+        if (student1.married) {
+          result = 1;
+        }
+
+        if (student2.married) {
+          result = -1;
+        }
+        break;
```
naziamoff, line 64: does not look right. You can use usual logical operators and it will work.

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+          - calculateAverageGrade(student2.grades);
+        break;
+
+      default:
+        throw new Error('Cannot sort that data type');
+    }
+
+    if (order === 'desc') {
+      result *= -1;
+    }
```
naziamoff, line 78: 1000 IQ move. I would make it a ternary though

image

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverageGrade(grades: number[]): number {
+  if (grades.length === 0) {
```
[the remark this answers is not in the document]
vladyslav-lapshyn: @naziamoff  I added to have extra validation if we our array doesn't have any grades. But you're right, I don't need this validation.

src/sortStudents.ts
vladyslav-lapshyn: @naziamoff thanks, it's a good point, I didn't think about ternary here

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+        result = student1[sortBy].localeCompare(student2[sortBy]);
+        break;
+
+      case SortType.Age:
+        result = student1.age - student2.age;
+        break;
+
+      case SortType.Married:
+        result = Number(student1[sortBy]) - Number(student2[sortBy]);
+        break;
```
naziamoff, line 46: is there any difference between sorting logic of those two?
If you have it as `sortBy` dealing with sorting by age - you can group age and married sorting

naziamoff (approved):
good job 👍

PR #1374, opened 2023-05-31
Develop

serentinos:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  const getAvg = (arrNum: number[]): number => (
```
naziamoff, line 28: would be nice to move to the separate `helpers` file

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+
+  sortedStudents.sort(
+    (student1: Student, student2: Student) => {
+      switch (sortBy) {
+        case SortType.Name:
+        case SortType.Surname:
+          return order === 'asc'
+            ? student1[sortBy].localeCompare(student2[sortBy])
+            : student2[sortBy].localeCompare(student2[sortBy]);
+        case SortType.Age:
```
naziamoff, line 40: would be nice to separate cases with different options by adding empty line in between

naziamoff (approved):
Nice job! A couple of suggestions though

PR #1376, opened 2023-05-31
add solution

vchervinko:


naziamoff (changes requested):
switch works better when we have a pre-defined set of options

PR #1377, opened 2023-05-31
solution

kirilshorin:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
+  return grades.length
```
naziamoff, line 21: what happens if you don't have such validation?

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+    ? grades.reduce((sum, grade) => sum + grade, 0) / grades.length
+    : 0;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyStudents: Student[] = [...students];
```
naziamoff, line 31: sounds like a name for a function, as functions start with a verb. But if we do it vice versa - it becomes clear we're dealing with a variable
```suggestion
  const studentsCopy: Student[] = [...students];
```

naziamoff (approved):
looks good 👍 
A couple of minor suggestions though

PR #1380, opened 2023-05-31
solution

snuber-w:


naziamoff (changes requested):
if we have pre-determined conditions - we use switch

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+export type SortOrder = 'asc' | 'desc';
+
+function averageGrade(grades:number[]): number {
```
naziamoff, line 19: I expect something like `3.5` will be stored inside a variable like that, but instead you have it as a function. Not clear

naziamoff (approved):
functions should have the names starting with a verb. It must be perfectly clear what the function does by just seeing its name

PR #1381, opened 2023-05-31
add task solution

EugeneSherepa:


naziamoff (changes requested):
let's have just a single sort and implement the whole sorting logic inside a single callback

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: string,
+): Student[] {
+  return [...students].sort((a, b) => {
+    switch (sortBy) {
+      case 'name':
+      case 'surname':
```
naziamoff, line 33: and so on.
Basically that's the purpose for using enums - they are far more reliable then just the string values
```suggestion
      case SortBy.Name:
      case SortBy.Surname:
```

naziamoff (approved):
awesome job 👍

PR #1382, opened 2023-05-31
Solution

Yuliia-kucheriavenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+    case SortType.Married: {
+      const married
+        = sortedStudents.filter((student) => student[sortBy]);
+      const unmarried
+        = sortedStudents.filter((student) => !student[sortBy]);
+
+      return order === 'asc'
+        ? [...unmarried, ...married]
+        : [...married, ...unmarried];
+    }
```
naziamoff, line 56: you can do that with a simple logical operator

naziamoff (changes requested):
let's implement a single sort and make the switch inside the callback

naziamoff (changes requested):
not fixed

image

src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      case SortType.Age:
+      case SortType.Married:
+        return order === 'asc'
+          ? Number(studentA[sortBy]) - Number(studentB[sortBy])
+          : Number(studentB[sortBy]) - Number(studentA[sortBy]);
+      default:
+        return 0;
```
33Orange, line 51: Use default cases for throw error.
```suggestion
        throw new Error('Incorrect sort type');
```

33Orange (approved):
Good job

PR #1383, opened 2023-05-31
Solution

MaksimKad:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrades(grades: number[]): number {
```
naziamoff, line 20:
```suggestion
function getAverageGrade(grades: number[]): number {
```

naziamoff (changes requested):
let's now have the logic inside a single `.sort` method callback
```
.sort((a, b) => {
   switch () {
     ....
   }
})
```

PR #1384, opened 2023-05-31
sort students

YeMyr:


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+    ? currentValue - nextValue
+    : nextValue - currentValue;
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy: Student[] = [...students].map((item) => ({ ...item }));
```
33Orange, line 43: You don't need deep copy here.

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+    case SortType.AverageGrade:
+      return studentsCopy.sort(
+        (currentStudent: Student, nextStudent: Student) => {
+          return getSortValue(
+            order,
+            getAverageGrade(currentStudent),
+            getAverageGrade(nextStudent),
+          );
+        },
+      );
```
33Orange, line 77: You can use this switch case inside `sort` callback and return only sorting condition in every case.
So you will have only 1 call to the sort method.

YeMyr: 👌 fixed

src/sortStudents.ts
YeMyr: 👌 fixed

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'averageGrade',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(student: Student): number {
```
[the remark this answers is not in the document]
YeMyr: 👌 fixed

naziamoff (changes requested):
this one is not yet fixed.
```
array.sort((a, b) => {
  switch (condition) {
    ...sorting logic for different cases
  }
})
```

image

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(grades: number[]): number {
+  return grades.reduce(
+    (gradesSum: number, grade: number) => {
+      return grade + gradesSum;
+    },
+  ) / grades.length;
+}
```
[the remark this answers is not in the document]
YeMyr: Ok, I've understood 
please,, check last commit

naziamoff (approved):
🔥

PR #1385, opened 2023-05-31
Solution

PanyaPrimal:


naziamoff (approved):
Good job 👍 
One suggestion:

let's make our code more abstract and group the cases in which we compare by similar properties, or with the same logic. e.g. strings, numbers/boolean etc.

PR #1392, opened 2023-06-01
solution, needs switch improvement

SofiaLiubchenko:


naziamoff (changes requested):
interesting approach, well done)

Let's now try to get rid of one switch and try to implement asc/desc logic inside the case

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+          ? Number(a[sortBy]) - Number(b[sortBy])
+          : Number(b[sortBy]) - Number(a[sortBy]);
+
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? getAverageGrade(a.grades) - getAverageGrade(b.grades)
+          : getAverageGrade(b.grades) - getAverageGrade(a.grades);
+
+      default:
+        throw new Error('Not valid data');
```
33Orange, line 51: Use more descriptive messages for errors

33Orange (approved):
Well done !

PR #1394, opened 2023-06-02
Students sorted

olehmahera:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+};
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const sortedStudents:Student[] = [...students];
```
naziamoff, line 31: they are not yet sorted, this is just students copy

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+  return sortedStudents.sort((firstStud, secondStud) => {
+    const studAField = firstStud[sortBy];
+    const studBField = secondStud[sortBy];
```
naziamoff, line 35: I guess it is more clear this way.
```suggestion
    const firstStudentProperty = firstStud[sortBy];
    const secondStudentProperty = secondStud[sortBy];
```

naziamoff (approved):
🔥

PR #1395, opened 2023-06-02
solution

thienlaoo:


src/sortStudents.ts
```diff
@@ -1,16 +1,48 @@
...
+      case SortType.Surname:
+        return order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+      case SortType.AverageGrade:
+        return order === 'asc'
+          ? calculateAverage(a[sortBy]) - calculateAverage(b[sortBy])
+          : calculateAverage(b[sortBy]) - calculateAverage(a[sortBy]);
+      default:
+        return 0;
```
33Orange, line 45: Use default case for error handling.
```suggestion
        throw new Error('Incorrect sort type')
```

PR #1402, opened 2023-06-10
sortStudent

Tetiana-Hishchak:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc'|'desc';
```
VitaliyBondarenko1982, line 19:
```suggestion
export type SortOrder = 'asc' | 'desc';
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+  const copyStudents:Student[] = [...students];
+
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
+    if (order === 'asc') {
+      return copyStudents.sort((a:Student,
+        b:Student) => a[sortBy].localeCompare(b[sortBy]));
+    }
+
+    return copyStudents.sort((a:Student,
+      b:Student) => b[sortBy].localeCompare(a[sortBy]));
```
VitaliyBondarenko1982, line 39: consider to use ternary
   here and in other cases
   
```suggestion
      return copyStudents.sort((a:Student,
        b:Student) => order === 'asc' 
          ? a[sortBy].localeCompare(b[sortBy]))
           : b[sortBy].localeCompare(a[sortBy]));
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
-  // write your function
+  return sum / grades.length;
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+) :Student[] {
+  const copyStudents:Student[] = [...students];
+
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
```
VitaliyBondarenko1982, line 32: consider to use switch case instead multiply if
```suggestion
  switch(sortBy) {
    case SortType.Name:
    case SortType.Surname:
    
       // logic and ohter cases
     default: 
       return students
  }
```

VitaliyBondarenko1982 (changes requested):

etojeDenys (approved):
lgtm

PR #1404, opened 2023-06-11
solution

Absolution26:


src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+  const copyStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (order === 'asc') {
+    switch (sortBy) {
+      case SortType.Name:
+        return copyStudents.sort((a, b) => a.name.localeCompare(b.name));
+      case SortType.Surname:
+        return copyStudents.sort((a, b) => a.surname.localeCompare(b.surname));
```
TarasHoliuk, line 33:
```suggestion
      case SortType.Name:
      case SortType.Surname:
        return copyStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
```
you are doing the same but with different keys
add values to `SortType` enum properties that match `Student` object properties to be able use it as key

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+      case SortType.Age:
+        return copyStudents.sort((a, b) => a.age - b.age);
+      case SortType.Married:
+        return copyStudents
+          .sort((a, b) => (a.married ? 1 : -1) - (b.married ? 1 : -1));
```
TarasHoliuk, line 38: const a = false;
const b = +a; // -1

you may transform boolean values into numbers and handle it in the same way as `Age`

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+      case SortType.AverageGrade:
+        return copyStudents
+          .sort((a, b) => (a.grades
+            .reduce((sum, currentGrade) => sum + currentGrade, 0)
```
TarasHoliuk, line 42: move grades calculation to a separate function since you use it ore than once

src/sortStudents.ts
```diff
@@ -1,16 +1,77 @@
...
+            / a.grades.length)
+          - (b.grades
+            .reduce((sum, currentGrade) => sum + currentGrade, 0)
+            / b.grades.length));
+      default:
+        break;
+    }
+  }
+
+  if (order === 'desc') {
```
TarasHoliuk, line 52: probably will look more laconic if you handle `order` in each separate case, but it's not critical - up to you to change it or not

TarasHoliuk (changes requested):
Well done :+1: 
I added a few suggestions on how to improve your code :)

src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+    case SortType.Name:
+    case SortType.Surname:
+      return isDesc
+        ? copyStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]))
+        : copyStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+    case SortType.Age:
+    case SortType.Married:
+      return isDesc
+        ? copyStudents.sort((a, b) => +b[sortBy] - +a[sortBy])
+        : copyStudents.sort((a, b) => +a[sortBy] - +b[sortBy]);
```
lerastarynets, line 43: The '+' before each student's property is not necessary. Remove the '+' signs.

lerastarynets (approved):
Well done) Keep up the great work!

src/sortStudents.ts
Absolution26: But I transformed boolean value into number. Anyway thank you for approve

PR #1411, opened 2023-06-13
solution

Gosyav:


src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const copyStudents: Student[] = [...students];
+  let sortedStudents: Student[] = copyStudents;
```
mykhalenych, line 27: No need to create a separate variable 'sortedStudents'; use 'copyStudents' directly for sorting.

src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  // eslint-disable-next-line default-case
+  switch (sortBy) {
```
mykhalenych, line 30: Add a default case that returns the original array.

mykhalenych: and better will be 
```suggestion
  copyStudents.sort((student1: Student, student2: Student) => {
    switch (sortBy) {...yourCases}
  })
```

src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
+    case SortType.Name: {
+      sortedStudents
+        = copyStudents.sort((student1: Student, student2: Student) => {
+          return order === 'asc'
+            ? student1.name.localeCompare(student2.name)
+            : student2.name.localeCompare(student1.name);
+        });
+
+      break;
```
mykhalenych, line 39: and in cases return value without break

mykhalenych (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
+          : student2AverageGrade - student1AverageGrade;
+      }
+
+      default: {
+        return 0;
+      }
+    }
+  });
+
+  return copyStudents;
```
mykhalenych, line 75:
```suggestion
```

mykhalenych (changes requested):

mykhalenych (approved):

PR #1417, opened 2023-06-13
added solution

vladyslav10shevchenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+        aValue = a.name;
+        bValue = b.name;
+        break;
+
+      case SortType.Surname:
+        aValue = a.surname;
+        bValue = b.surname;
+        break;
```
etojeDenys, line 45:
```suggestion
      case SortType.Name:
      case SortType.Surname:
        aValue = a[sortBy];
        bValue = b[sortBy];
        break;
```
we can do the same with age and avg grade

etojeDenys (changes requested):

Viktor-Kost (approved):

PR #1418, opened 2023-06-14
Solution

offsset:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  students: Student [],
+  sortBy: SortType,
+  order: SortOrder,
+):Student [] {
+  const studentsNew: Student [] = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case 'age': {
```
OleziO, line 28: Move all cases to the enum

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      const result: Student[] = studentsNew.sort((x, y) => {
+        return (order === 'asc' ? x[sortBy] - y[sortBy]
+          : y[sortBy] - x[sortBy]);
+      });
+
+      return result;
+    }
+
+    case 'name':
+    /* falls through */
```
OleziO, line 38: Remove all comments from the code.

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      });
+
+      return result;
+    }
+
+    case 'grades': {
+      const result: Student[] = studentsNew.sort((x, y) => {
+        return (
+          order === 'asc'
+            ? (x[sortBy].reduce((a, summ) => a + summ)) / x[sortBy].length
```
OleziO, line 53: Move calculation to the variable to avoid code duplication

OleziO (changes requested):

src/sortStudents.ts
sTorba24: don't see `maried` case

sTorba24: also use values from `enum`

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+      return result;
+    }
+
+    case 'name':
+
+    // eslint-disable-next-line no-fallthrough
+    case 'surname': {
+      const result: Student[] = studentsNew.sort((x, y) => {
+        return (order === 'asc' ? x[sortBy].localeCompare(y[sortBy])
+          : y[sortBy].localeCompare(x[sortBy]));
```
sTorba24, line 43:
```suggestion
        return (order === 'asc' 
          ? x[sortBy].localeCompare(y[sortBy])
          : y[sortBy].localeCompare(x[sortBy])
        );
```

sTorba24 (changes requested):
Almost good! Lets improve it

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+    case SortType.Married: {
+      const result: Student[] = studentsNew.sort((x, y) => {
+        return (order === 'asc' ? Number(x[sortBy]) - Number(y[sortBy])
+          : Number(y[sortBy]) - Number(x[sortBy]));
+      });
+
+      return result;
+    }
+
+    case SortType.Age: {
```
VitaliyBondarenko1982, line 37: For Married and Age same logic. they can be united
```
case SortType.Married:
case SortType.Age: 
  // do something

VitaliyBondarenko1982 (approved):

PR #1419, opened 2023-06-14
solution

DanyloPasko:


volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
 
 export interface Student {
-  // describe Student interface
+  name: string,
```
volodymyr-soltys97, line 3: Properties in an interface should be delimited by semicolons, not commas.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1422, opened 2023-06-14
add task solution

sasdaiv:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const asc = order === 'asc';
 
+  const callBack = (a: Student, b: Student): number => {
```
etojeDenys, line 34: Rename 'callBack' to 'compareFunction' for better readability and to follow naming conventions.

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grade',
+}
+
+function getAverageGrade(grades: number[]): number {
+  return grades.length
+    ? grades.reduce((sum: number, n: number) => (sum + n)) / grades.length
+    : 0;
```
etojeDenys, line 21: let's simplify it
```suggestion
  return grades.reduce((sum: number, n: number) => (sum + n), 0) / grades.length;
```

etojeDenys (changes requested):

PR #1424, opened 2023-06-14
task solved

Nazarht:


src/sortStudents.ts
```diff
@@ -1,16 +1,114 @@
...
-
-export function sortStudents(students, sortBy, order) {
+export function sortStudents(
+  students: Student[],
+  sortBy:SortType,
+  order: SortOrder,
+): Student[] {
   // write your function
+  const newArr: Student[] = [...students];
+
```
anastasiiavorobiova, line 30: Consider renaming 'newArr' to 'sortedStudents' to better describe its purpose.

src/sortStudents.ts
```diff
@@ -1,16 +1,114 @@
...
+  newArr
+    .sort((studentOne: Student, studentTwo: Student): number => {
+      switch (sortBy) {
+        case SortType.Name:
+          if (order === 'asc') {
```
anastasiiavorobiova, line 35: It's a magic string, consider using constants for this with a meaningful name

src/sortStudents.ts
```diff
@@ -1,16 +1,114 @@
...
+            return studentOne.name.toLowerCase() > studentTwo.name.toLowerCase()
```
anastasiiavorobiova, line 36: it's better to use localeCompare for sorting strings to avoid unexpected results

anastasiiavorobiova (changes requested):
Great start! Let's make it more perfect

src/sortStudents.ts
Nazarht: Didn't understand what exactly do you mean

src/sortStudents.ts
```diff
@@ -1,16 +1,112 @@
 
 export interface Student {
   // describe Student interface
```
OleziO, line 3: Remove all comments

src/sortStudents.ts
OleziO: Rename to `sortedStudents`. Its more meaningful name

src/sortStudents.ts
OleziO: Move 'asc' to the variable and use it.

OleziO (changes requested):

OleziO (changes requested):
Fix all coments

Nazarht:
What do you mean by 'all comments'? I don't have any comments in my code. And all the previous requests I changed.

PR #1433, opened 2023-06-19
sort function created

JasterHim:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+      case SortType.Age:
+        return (order === 'asc')
+          ? +first[sortBy] - +second[sortBy]
+          : +second[sortBy] - +first[sortBy];
+
+      case SortType.Married:
+        return (order === 'asc')
+          ? +first[sortBy] - +second[sortBy]
+          : +second[sortBy] - +first[sortBy];
```
sTorba24, line 52: as it's the same you can use multiple cases
```suggestion
      case SortType.Married:
      case SortType.Age:
        return (order === 'asc')
          ? +first[sortBy] - +second[sortBy]
          : +second[sortBy] - +first[sortBy];
```

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+    switch (sortBy) {
+      case SortType.Name:
+        return (order === 'asc')
+          ? first[sortBy].localeCompare(second[sortBy])
+          : second[sortBy].localeCompare(first[sortBy]);
+
+      case SortType.Surname:
+        return (order === 'asc')
+          ? first[sortBy].localeCompare(second[sortBy])
+          : second[sortBy].localeCompare(first[sortBy]);
```
sTorba24, line 42: same here

sTorba24 (changes requested):
Almost good! Lets improve it :)

sTorba24 (approved):
Good work!

PR #1442, opened 2023-06-23
Add students sort

AmadeuszAndroid:


src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+    case SortType.Age:
+      return sortedStudents.sort((s1, s2) => {
+        return order === 'asc'
+          ? s1.age - s2.age
+          : s2.age - s1.age;
+      });
+    case SortType.Married:
+      return sortedStudents.sort((s1, s2) => {
+        return order === 'asc'
+          ? +s1.married - +s2.married
```
darokrk, line 61: Instead of using here an implicit coercion types by using `+, -` for better readability, I would suggest use here a `Number(s1.married) - Number(s2.married)` constructor, then it's more clear and we have sure that other programmer would know what exactly we want to reach in our code :)

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+  Married,
+  AverageGrade
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(student: Student): number {
+  const grades = [...student.grades];
```
darokrk, line 22: Think you don't need to do that here, the `reduce` method does not modifies an original array :)

darokrk (changes requested):
Looks good, small improvements need and we have it done 💪🏽

src/sortStudents.ts
AmadeuszAndroid: I switched to `Number`.

src/sortStudents.ts
AmadeuszAndroid: I changed copy to dectructuring.

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        return order === 'asc'
+          ? s1.name.localeCompare(s2.name)
+          : s2.name.localeCompare(s1.name);
+      });
+    case SortType.Surname:
+      return sortedStudents.sort((s1, s2) => {
+        return order === 'asc'
+          ? s1.surname.localeCompare(s2.surname)
+          : s2.surname.localeCompare(s1.surname);
+      });
```
DorotaLeniecDev, line 51: I think you could combine this two cases like:

```suggestion
    case SortType.Name:
    case SortType.Surname:
      return sortedStudents.sort((s1, s2) => {
        return order === 'asc'
          ? s1[sortBy].localeCompare(s2[sortBy])
          : s2[sortBy].localeCompare(s1[sortBy]);
      });
```

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        return order === 'asc'
+          ? s1.age - s2.age
+          : s2.age - s1.age;
+      });
+    case SortType.Married:
+      return sortedStudents.sort((s1, s2) => {
+        return order === 'asc'
+          ? Number(s1.married) - Number(s2.married)
+          : Number(s2.married) - Number(s1.married);
+      });
```
DorotaLeniecDev, line 63: Same here:
```suggestion
    case SortType.Married:
    case SortType.Age:
      return sortedStudents.sort((s1, s2) => {
        return order === 'asc'
          ? Number(s1[sortBy]) - Number(s[sortBy]))
          : Number(s2[sortBy]) - Number(s1[sortBy]));
      });
```

DorotaLeniecDev (changes requested):
Good job, small improvement suggestion

src/sortStudents.ts
AmadeuszAndroid: I merged these cases.

src/sortStudents.ts
AmadeuszAndroid: I merged these cases.

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverageGrade(student: Student): number {
```
Radoslaw-Czerniawski, line 21: This could be shortened and made more generic:
image
- this function unnecessarily received whole student object despite only working on it's grades, now it has less responsibility and is more generic
- there is no need for conditions, with starting val of 0 reduce in this case will return 0 for empty array

Radoslaw-Czerniawski (approved):
Accepted, looks very good :D, left one small refactor suggestion

PR #1444, opened 2023-06-24
Solution

Nataliia95254:


src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+  if (grades.length === 0) {
+    return 0;
+  }
+
+  const sum = grades.reduce((acc, grade) => acc + grade, 0);
+
+  return sum / grades.length;
 }
 
 // create SortOrder type
```
OleziO, line 28: Remove all comments from the code.

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+
+    case SortType.Surname:
+      return result.sort(
+        (a: Student, b: Student) => {
+          if (order === 'desc') {
+            b.surname.localeCompare(a.surname);
+          }
+
+          return a.surname.localeCompare(b.surname);
+        },
```
OleziO, line 59: Combine them to avoid code duplication.

OleziO (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+  :[] {
+  const result: [] = JSON.parse(JSON.stringify(students));
+
+  switch (sortBy) {
+    case SortType.Name:
+      return sortString(result, 'name', order);
+
+    case SortType.Surname:
+      return sortString(result, 'surname', order);
+
```
etojeDenys, line 60: we can group these cases
```suggestion
    case SortType.Name:
    case SortType.Surname:
      return sortString(result, sortBy, order);

```

src/sortStudents.ts
```diff
@@ -1,16 +1,104 @@
...
+            return -1;
+          }
+
+          if (!a.married && b.married) {
+            return 1;
+          }
+
+          return 0;
+        },
+      );
```
etojeDenys, line 85: same here
```suggestion
    case SortType.Age:
    case SortType.Married:
      return result.sort(
        (a: Student, b: Student) => {
          if (order === 'desc') {
            return +b[sortBy] - +a[sortBy];
          }

          return +a[sortBy]- +b[sortBy];
        },
      );

    
     
```

etojeDenys (changes requested):
also don't forget to update enum

VitaliyBondarenko1982 (approved):

PR #1446, opened 2023-06-24
solution

vlad-fil:


src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+  }
+
+  if (sortBy === SortType.Married && order === 'asc') {
+    result.sort((a: Student, b: Student) => +a.married - +b.married);
+  } else if (sortBy === SortType.Married && order === 'desc') {
+    result.sort((a: Student, b: Student) => +b.married - +a.married);
+  }
+
+  if (sortBy === SortType.AverageGrade && order === 'asc') {
+    result.sort((a: Student, b: Student) => (
```
etojeDenys, line 53: Extract the logic of calculating the average grade into a separate function to improve readability and maintainability.

src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+  } else if (sortBy === SortType.Surname && order === 'desc') {
+    result.sort((a: Student, b: Student) => b.surname.localeCompare(a.surname));
+  }
+
+  if (sortBy === SortType.Married && order === 'asc') {
+    result.sort((a: Student, b: Student) => +a.married - +b.married);
+  } else if (sortBy === SortType.Married && order === 'desc') {
+    result.sort((a: Student, b: Student) => +b.married - +a.married);
+  }
+
```
etojeDenys, line 51: let's use switch\case instead of multiple if/else

etojeDenys (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+    case SortType.Name:
+      return (order === 'asc')
+        ? result.sort((a, b) => a.name.localeCompare(b.name))
+        : result.sort((a, b) => b.name.localeCompare(a.name));
+
+    case SortType.Surname:
+      return (order === 'asc')
+        ? result.sort((a, b) => a.surname.localeCompare(b.surname))
+        : result.sort((a, b) => b.surname.localeCompare(a.surname));
```
etojeDenys, line 47: we can combine some cases
```suggestion
    case SortType.Name:
    case SortType.Surname:
      return (order === 'asc')
        ? result.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
        : result.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```

(don't forget to update the enum)

etojeDenys (changes requested):

PR #1449, opened 2023-06-27
add task solution

denys2:


src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[], sortBy: SortType,
+  order: SortOrder,
+): Student[] {
```
polosanya, line 22:
```suggestion
export function sortStudents(
  students: Student[], 
  sortBy: SortType,
  order: SortOrder,
): Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+  // Create a new array to hold the sorted students
+  const sortedStudents = [...students];
```
polosanya, line 24:
```suggestion
  const studentsCopy = [...students];
```
Actually, they are not sorted, just a copy

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+
+  // Define the sorting logic based on the provided sortBy parameter
+  switch (sortBy) {
+    case SortType.Name:
+      sortedStudents.sort((a, b) => a.name.localeCompare(b.name));
+      break;
+    case SortType.Surname:
+      sortedStudents.sort((a, b) => a.surname.localeCompare(b.surname));
```
polosanya, line 32: In both cases you do the same, consider combining these two cases
image

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+      break;
+    case SortType.Age:
+      sortedStudents.sort((a, b) => a.age - b.age);
+      break;
+    case SortType.Married:
+      sortedStudents.sort((a, b) => {
```
polosanya, line 38: Same here

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+        if (a.married === b.married) {
+          return 0;
+        }
+
+        if (order === 'asc') {
+          return a.married ? -1 : 1;
+        // eslint-disable-next-line no-else-return
+        } else {
+          return a.married ? 1 : -1;
+        }
```
polosanya, line 48:
```suggestion
        if (order === 'asc') {
          return a.married ? -1 : 1;
        }
        
        return a.married ? 1 : -1;
```
Don't disable the eslint, but try to fix its problem. `else` is redundant here as code will not go further after return anyway

src/sortStudents.ts
```diff
@@ -1,16 +1,79 @@
...
+      });
+      break;
+    case SortType.AverageGrade:
+      sortedStudents.sort((a, b) => {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        const averageGradeA = calculateAverageGrade(a);
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
polosanya, line 55: Fix the problem, not disable the eslint

polosanya (changes requested):
Good job! Let's improve your code a bit.
Fix all eslint problems. You can also remove all comments as it's clear from your code anyway

Anastasiia-Svintsova (approved):

PR #1451, opened 2023-06-28
Solution

natalia-maryshko:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function getAverage(arr: number[]): number {
+  return arr.reduce((a: number, b: number) => a + b) / arr.length;
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+) : Student[] {
```
polosanya, line 27:
```suggestion
export function sortStudents(
  students: Student[], 
  sortBy: SortType, 
  order: SortOrder,
) : Student[] {
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  const result = [...students];
```
```suggestion
  const studentsCopy = [...students];
```
It is better to avoid such names as 'result'. Try to provide meaningful and descriptive names

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+
+  switch (sortBy) {
+    case 'name':
+    case 'surname':
```
polosanya, line 32: Use Enum values instead of hardcoded

polosanya (changes requested):
Almost done!

qoa11a (approved):
Good job! 👍

PR #1459, opened 2023-07-03
Solution

Yulia2806:


src/sortStudents.ts
```diff
@@ -1,16 +1,53 @@
...
+      case SortType.Age:
+        return order === 'asc'
+          ? Number(a.age) - Number(b.age)
+          : Number(b.age) - Number(a.age);
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+      case SortType.Married:
+        return order === 'asc'
+          ? Number(a.married) - Number(b.married)
```
lerastarynets, line 45: Booleans should not be converted to numbers for comparison. Instead, use JavaScript's native comparison operators.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

lerastarynets (approved):
All requirements are met. You're doing a fantastic job, keep going!

PR #1463, opened 2023-07-03
solution

vshepel:


src/sortStudents.ts
```diff
@@ -1,16 +1,41 @@
...
+): Student[] {
+  const sortOrder: number = order === 'asc' ? 1 : -1;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return [...students].sort((a, b) => {
+    switch (sortBy) {
+      case SortType.Age:
+      case SortType.Married:
+        return sortOrder * (Number(a[sortBy]) - Number(b[sortBy]));
```
IvanFesenko, line 34: this logic is not clear
use ternary operator insteam magic multiple
Fix all cases
```suggestion
      case SortType.Married:
        return sortOrder ==='asc'
         ? Number(a[sortBy]) - Number(b[sortBy])
         : Number(b[sortBy]) - Number(a[sortBy])
```

IvanFesenko (changes requested):
Good job. Check my comment

IvanFesenko (approved):

PR #1470, opened 2023-07-04
add task solution

Vayts:


roman-mirzoian (approved):
A job well done, but try to break large chunks of code into smaller chunks of code that are easier to read and extend functionally.
Also pay attention to the principles of programming:
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

PR #1475, opened 2023-07-05
add task solution

kbekher:


src/sortStudents.ts
```diff
@@ -1,16 +1,48 @@
+// import { type } from 'os';
```
Viktor-Kost, line 1: There's an unused import statement that's commented out. If it's not needed, it's better to remove it completely to avoid clutter.

Viktor-Kost (approved):
🔥

PR #1477, opened 2023-07-06
Task done

AntonBaranskyi:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+        ? copyStudents.sort((a, b) => +a.married - +b.married)
+        : copyStudents.sort((a, b) => +b.married - +a.married);
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? copyStudents.sort(
+          (a, b) => findAverageGrade(a.grades) - findAverageGrade(b.grades),
+        )
+        : copyStudents.sort(
+          (a, b) => findAverageGrade(b.grades) - findAverageGrade(a.grades),
+        );
```
YegorKorenkov, line 55: You can firstly sort by second param(name/surname/...) and after that sort by asc/desc instead of use ternary operator in each case

YegorKorenkov (changes requested):

YegorKorenkov (approved):
Approve, sorry misunderstood the assignment

PR #1490, opened 2023-07-07
solution

ovynarchuk:


src/sortStudents.ts
```diff
@@ -1,16 +1,54 @@
...
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function calculateAverageGrade({ grades }: Student): number {
+  return grades.reduce((average, current) => (
+    average + current)) / grades.length;
+}
```
qoa11a, line 23: Instead of passing the entire Student object here, you could pass just a grades property.

qoa11a (approved):

PR #1498, opened 2023-07-07
add task solution

NazarBaraban:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      return order === 'asc'
+        ? sortedStudents.sort((a, b) => a.name.localeCompare(b.name))
+        : sortedStudents.sort((a, b) => b.name.localeCompare(a.name));
+    case SortType.Surname:
+      return order === 'asc'
+        ? sortedStudents.sort((a, b) => a.surname.localeCompare(b.surname))
+        : sortedStudents.sort((a, b) => b.surname.localeCompare(a.surname));
```
And678, line 40: Checklist: `[CODE STYLE] - DRY. Combine logic for sorting by fields with same data types.`

```suggestion
  const sortOrder: number = order === 'asc' ? 1 : -1;
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return sortedStudents.sort((a, b) => sortorder * a[sortBy].localeCompare(b[sortBy]));
```

Do the same for Age and Married

And678 (changes requested):
GJ, but the code can be a bit DRYer :)

PR #1503, opened 2023-07-07
Solution

outofs:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+      switch (sortBy) {
+        case SortType.Name:
+          return order !== 'desc'
+            ? stud1.name.localeCompare(stud2.name)
+            : stud2.name.localeCompare(stud1.name);
+
+        case SortType.Surname:
+          return order !== 'desc'
+            ? stud1.surname.localeCompare(stud2.surname)
+            : stud2.surname.localeCompare(stud1.surname);
```
And678, line 43: Checklist. `[CODE STYLE] - DRY. Combine logic for sorting by fields with same data types.` Same for other cases
```suggestion
    const sortOrder: number = order === 'asc' ? 1 : -1;
    copyStudents.sort((stud1: Student, stud2: Student) => {
      switch (sortBy) {
        case SortType.Name:
        case SortType.Surname:
          return sortOrder * stud1[sortBy].localeCompare(stud2[sortBy]);
```

And678 (changes requested):
Good start, but needs to be DRYed up a bit

qoa11a (approved):

PR #1506, opened 2023-07-07
solution

AnatoliiLykhohodenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  const sortedStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? sortedStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : sortedStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```
And678, line 35: Same can be done for other cases
```suggestion
  const sortOrder = order === 'asc' ? 1 : -1;
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return sortedStudents.sort((a, b) => sortOrder* a[sortBy].localeCompare(b[sortBy]));
```

And678 (changes requested):
One thing to improve and this is done :100:

anstsot (approved):
approving:sparkles:

PR #1507, opened 2023-07-07
add task solution

klmyana:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+    case SortType.Surname:
+      if (order === 'asc') {
+        return sortedStudents.sort(
+          (st1, st2) => st1[sortBy].localeCompare(st2[sortBy]),
+        );
+      }
+
+      return sortedStudents.sort(
+        (st1, st2) => st1[sortBy].localeCompare(st2[sortBy]),
+      );
```
And678, line 43: Can be shorter here. Also do this for all of the `case`s
```suggestion
  const sortOrder = order === 'asc' ? 1 : -1;
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return sortedStudents.sort(
        (st1, st2) => sortOrder * st1[sortBy].localeCompare(st2[sortBy]),
      );
```

And678 (changes requested):
GJ, only one thing to imrove:

PR #1510, opened 2023-07-08
writed code

Bonum-taurus:


src/sortStudents.ts
```diff
@@ -1,16 +1,132 @@
+// import { defaultCipherList } from 'constants';
```
volodymyr-soltys97, line 1: The import statement is commented out. If it's not being used, it's better to remove it to keep the code clean.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1516, opened 2023-07-09
add solution

dimarogkov:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+      studentArr.sort((a: Student, b: Student) => (order === 'asc'
+        ? a.name.localeCompare(b.name)
+        : b.name.localeCompare(a.name)));
+      break;
+
+    case SortType.Surname:
+      studentArr.sort((a: Student, b: Student) => (order === 'asc'
+        ? a.surname.localeCompare(b.surname)
+        : b.surname.localeCompare(a.surname)));
+      break;
```
VitaliyBondarenko1982, line 41:
```suggestion
    case SortType.Name:
    case SortType.Surname:
      studentArr.sort((a: Student, b: Student) => (order === 'asc'
        ? a[sortBy].localeCompare(b[sortBy])
        : b[sortBy].localeCompare(a[sortBy])));
      break;
```
this to cases with same logic

src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+          return 0;
+        }
+
+        if ((a.married && order === 'asc') || b.married) {
+          return 1;
+        }
+
+        return -1;
+      });
+      break;
```
VitaliyBondarenko1982, line 61:
```suggestion
    case SortType.Age:
    case SortType.Married:
      studentArr.sort((a: Student, b: Student) => (order === 'asc'
        ? +a[sortBy] - +b[sortBy]
        : +b[sortBy] - +a[sortBy]));
      break;
```
same for age and married
converting true/false to number return 1 or 0

VitaliyBondarenko1982 (changes requested):

src/sortStudents.ts
dimarogkov: I have some errors if I do like you
image
image

How I can fix this problem?

etojeDenys (changes requested):
feel free to ask in chat

VitaliyBondarenko1982 (approved):

PR #1517, opened 2023-07-09
Solution

kylyvnyk:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
+      return order === 'asc'
+        ? copyStudents.sort((a, b) => a.name.localeCompare(b.name))
+        : copyStudents.sort((a, b) => b.name.localeCompare(a.name));
+    case SortType.Surname:
+      return order === 'asc'
+        ? copyStudents.sort((a, b) => a.surname.localeCompare(b.surname))
+        : copyStudents.sort((a, b) => b.surname.localeCompare(a.surname));
```
And678, line 40: Checklist. Avoid duplication in these cases.
```suggestion
  const sortOrder = order === 'asc' ? 1 : -1;
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return sortOrder * copyStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
```

And678: Same for the other cases

And678 (changes requested):
Gj, some improvements below

PR #1521, opened 2023-07-09
add task solution Ts_sort_students

TamaraShukalo:


YegorKorenkov (approved):
Good job!

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+    let result: number = 0;
+
+    switch (sortBy) {
+      case SortType.Name:
+        result = student1.name.localeCompare(student2.name);
+        break;
+
+      case SortType.Surname:
+        result = student1.surname.localeCompare(student2.surname);
+        break;
```
And678, line 41: Check the checklist. You can combine these. Also the same applies to Age and Married
```suggestion
      case SortType.Name:
      case SortType.Surname:
        result = student1[sortBy].localeCompare(student2[sortBy]);
        break;
```

And678 (changes requested):
Gj, only one thing to improve here:

anstsot (approved):
Good work, approving:sparkles:

PR #1522, opened 2023-07-09
add task solution

ArtemBehter:


src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] | number {
```
darokrk, line 27: The function sortStudents should always return an array of students, returning a number isn't appropriate. If no sort condition is met, just return the unchanged array.

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+  switch (sortBy) {
+    case SortType.Name:
+    case SortType.Surname:
+      return order === 'asc'
+        ? allStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
+        : allStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
+    case SortType.Age:
+    case SortType.Married:
+      return order === 'asc'
+        ? allStudents.sort((a, b) => +a[sortBy] - +b[sortBy])
```
darokrk, line 43: Please use explicit conversion to number type by using `Number()` constructor, it is more readable e.g.

`Number(a[sortBy]) - Number(b[sortBy])`

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+        : allStudents.sort((a, b) => +b[sortBy] - +a[sortBy]);
```
darokrk, line 44: same as above

src/sortStudents.ts
```diff
@@ -1,16 +1,58 @@
...
+        ? allStudents.sort((a, b) => (
+          findAverageGrade(a[sortBy])
+          - findAverageGrade(b[sortBy])
+        ))
+        : allStudents.sort((a, b) => (
+          findAverageGrade(b[sortBy])
+          - findAverageGrade(a[sortBy])
+        ));
+    default:
+      return 0;
```
darokrk, line 56: Think this should just return the unchanged array instead of number

darokrk (changes requested):
Few suggestions to fix and we got this 💪🏽

PR #1524, opened 2023-07-09
Implement sortStudents function

taras-bilas:


src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+          (firstStudent, nextStudent) => firstStudent[sortBy]
+            .localeCompare(nextStudent[sortBy]),
+        );
+      } else {
+        sortedStudents = sortedStudents.sort(
+          (firstStudent, nextStudent) => nextStudent[sortBy]
+            .localeCompare(firstStudent[sortBy]),
+        );
+      }
+      break;
```
And678, line 38: Can be shortened, same for the rest of cases
```suggestion
  const sortOrder = order === 'asc' ? 1 : -1;
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      sortedStudents = sortedStudents.sort(
        (firstStudent, nextStudent) => sortOrder * firstStudent[sortBy]
          .localeCompare(nextStudent[sortBy]),
      );
      break;
```

And678 (changes requested):
GJ, but can be improved

anstsot (approved):
Good work, approving:sparkles:

PR #1527, opened 2023-07-10
Solution

artyomwhite:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+
+        return firstStudent.age - secondStudent.age;
+
+      case SortType.Married:
+        if (order === 'desc') {
+          return Number(secondStudent.married) - Number(firstStudent.married);
+        }
+
+        return Number(firstStudent.married) - Number(secondStudent.married);
+
```
etojeDenys, line 54:
```suggestion
      case SortType.Age:
      case SortType.Married:
        if (order === 'desc') {
          return +secondStudent[sortBy] - +firstStudent[sortBy];
        }

        return +firstStudent[sortBy] - +secondStudent[sortBy];

```
we can group these cases too

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #1533, opened 2023-07-12
add task solution

vetal-hovenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+        averageGradeOfFirst = student1.grades
+          .reduce((prev, grade) => prev + grade, 0) / student1.grades.length;
+
+        averageGradeOfSecond = student2.grades
+          .reduce((prev, grade) => prev + grade, 0) / student2.grades.length;
+
+        return order === 'asc' ? averageGradeOfFirst - averageGradeOfSecond
+          : averageGradeOfSecond - averageGradeOfFirst;
+
+      default:
```
anastasiiavorobiova, line 59: In the default case, returning 0 can cause unexpected behavior if the SortType is not recognized. It's better to throw an error or handle this case properly.

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+        return order === 'asc'
+          ? student1.surname.localeCompare(student2.surname)
+          : student2.surname.localeCompare(student1.surname);
+
+      case SortType.Age:
+        return order === 'asc' ? student1.age - student2.age
+          : student2.age - student1.age;
+
+      case SortType.Married:
+        return order === 'asc' ? +student1.married - (+student2.married)
```
anastasiiavorobiova, line 46:
```suggestion
        return order === 'asc' ? Number(student1.married) - Number(student2.married)
```
Consider making type coercion clear and easily readable

anastasiiavorobiova (changes requested):
Great work! Added tiny comment so fix

anastasiiavorobiova (approved):
Great solution

PR #1534, opened 2023-07-12
solution

Vepr984:


src/sortStudents.ts
```diff
@@ -1,16 +1,52 @@
...
+  order: SortOrder,
+): Student[] {
+  const sortedStudents = students.map((student) => {
+    let averageGrade = 0;
+
+    if (student.grades.length > 0) {
+      averageGrade = student.grades.reduce(
+        (acc, grade) => acc + grade, 0,
+      ) / student.grades.length;
+    }
```
etojeDenys, line 32: cool idea but we don't want to change Student interface and change the initial array of students
also we don't want to make this calculation if sortBy is not a grades

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #1536, opened 2023-07-13
SortStudents

Dmytro-Tymchenko:


src/sortStudents.ts
```diff
@@ -1,16 +1,57 @@
...
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
+}
+
+export type SortOrder = 'asc' | 'desc';
+
+export function calcAverageGrade(grades: number[]): number {
```
mykhalenych, line 19: The reduce function should take the total as the first parameter and the grade as the second. It should be like this: return grades.reduce((total, grade) => total + grade) / grades.length;

mykhalenych (approved):

PR #1537, opened 2023-07-13
add solution

denis-kononenko:


anastasiiavorobiova (changes requested):
Nice work! Added tiny comments to fix

vlad-demchuk (approved):

PR #1538, opened 2023-07-14
Solution

yurii-shkrobut-m:


etojeDenys (changes requested):
think how to combine married and age cases

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+
+      case SortType.Age:
+      case SortType.Married:
+        return (
+          +studentA[sortBy] - +studentB[sortBy]
+        ) * modifier;
+
+      case SortType.AverageGrade:
+        return (getAverageGrade(studentA.grades)
+          - getAverageGrade(studentB.grades)) * modifier;
```
VitaliyBodnar, line 47: don't need to duplicate '* modifier' in all cases, better to do it once, after the switch case block

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1542, opened 2023-07-16
add task solution

reseneweb:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  const copyStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      copyStudents.sort((a, b) => a.name.localeCompare(b.name));
+      break;
+    case SortType.Surname:
+      copyStudents.sort((a, b) => a.surname.localeCompare(b.surname));
```
polosanya, line 32: In both cases the logic is the same. Consider combining them
image

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      break;
+
+    case SortType.Age:
+      copyStudents.sort((a, b) => b.age - a.age);
+      break;
+
+    case SortType.Married:
+      copyStudents.forEach((x) => ((x.married) ? 1 : 0));
+      copyStudents.sort((a, b) => Number(b.married) - Number(a.married));
+      copyStudents.forEach((x) => ((x.married) ? 'married' : ''));
```
polosanya, line 42: Do you really need three method calls to sort it? Let's try to do it only with sort method

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+          const bAverage
+           = b.grades.reduce((x, y) => x + y, 0) / b.grades.length;
+          const aAverage
+           = a.grades.reduce((x, y) => x + y, 0) / a.grades.length;
+
+          return bAverage - aAverage;
+        });
+      } else {
+        copyStudents.sort((a, b) => {
+          const bAverage
```
polosanya, line 57: Let's create a separate function for calculating average grades to make your code a bit cleaner

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averGrades(copyStudents: Student[], order: 'asc' | 'desc'): Student[] {
```
lerastarynets, line 20: Pls simplify this function, no need to check for `order` inside
You can check `order` in every `case` and make use of ternary operators

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+    case SortType.Surname:
+      copyStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
+      break;
+
+    case SortType.Age:
+      copyStudents.sort((a, b) => b[sortBy] - a[sortBy]);
+      break;
+
+    case SortType.Married:
+      copyStudents.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
```
polosanya, line 38:
```suggestion
    case SortType.Age:
    case SortType.Married:
      copyStudents.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]));
```
You can combine these cases too, as they are almost the same

src/sortStudents.ts
```diff
@@ -1,16 +1,55 @@
...
+      break;
+
+    case SortType.AverageGrade:
+      copyStudents.sort((a, b) => {
+        const bAverage = b.grades.reduce((x, y) => x + y, 0) / b.grades.length;
+        const aAverage = a.grades.reduce((x, y) => x + y, 0) / a.grades.length;
+
+        return order === 'desc' ? bAverage - aAverage : aAverage - bAverage;
```
polosanya, line 46: What about the order in other cases? Now it works only for average grade case

polosanya (changes requested):
Almost done!

witflash (approved):
Well done 👍

PR #1545, opened 2023-07-17
solution

IraPolchanin:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  function getAverageGrade(student: Student): number {
+    return student.grades.reduce((prev, cur) => prev + cur, 0)
+    / student.grades.length;
+  }
+
+  copyStudents.sort((firstStudent, secondStudent) => {
+    let a = firstStudent;
```
vlad-demchuk, line 35: There's no need to assign firstStudent and secondStudent to new variables a and b. You can use them directly in your code.

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+    let b = secondStudent;
+
+    if (order === 'desc') {
+      b = firstStudent;
```
vlad-demchuk, line 39: Instead of reassigning a and b depending on the order, you could determine the order in your return statements in the switch case. This would make your code more efficient.

vlad-demchuk (approved):

PR #1553, opened 2023-07-24
sortStudents() was almost implemented but tests student.age dont pass

oleksandrshtonda:
Тести на вік та married проходять тільки тоді, коли я запускаю їх власноруч. Для мене дивно, чому при однакових вхідних даних результати відрізняються :/

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,94 @@
-
 export interface Student {
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const orderNum: number = order === 'asc' ? 1 : -1;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  return students.sort((student1: Student, student2: Student): number => {
+    switch (sortBy) {
+      case SortType.Name:
+        if (student1.name.toLowerCase() > student2.name.toLowerCase()) {
+          return 1 * orderNum;
+        }
+
+        if (student1.name.toLowerCase() < student2.name.toLowerCase()) {
+          return -1 * orderNum;
+        }
+
+        return 0;
+
+      case SortType.Surname:
+        if (student1.surname.toLowerCase() > student2.surname.toLowerCase()) {
+          return 1 * orderNum;
+        }
+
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,94 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const orderNum: number = order === 'asc' ? 1 : -1;
+
+  return students.sort((student1: Student, student2: Student): number => {
```
volodymyr-soltys97, line 26: The sort method directly modifies the array it is called on. This doesn't fit the task requirements, which specify that the original array should not change. You should make a copy of the original array before sorting it.

src/sortStudents.ts
```diff
@@ -1,16 +1,94 @@
...
+        }
+
+        if (student1.age < student2.age) {
+          return -1 * orderNum;
+        }
+
+        return 0;
+
+      case SortType.Married:
+        if (student1.married > student2.married) {
```
volodymyr-soltys97, line 62: You're comparing boolean values as if they were numbers. This may lead to unexpected results. Instead, you should explicitly convert booleans to numbers (e.g., using Number() function) before comparison.

src/sortStudents.ts
```diff
@@ -1,16 +1,94 @@
...
+        }
+
+        if (student1.married < student2.married) {
+          return -1 * orderNum;
+        }
+
+        return 0;
+
+      case SortType.AverageGrade: {
+        const student1AverageGrade = student1.grades.reduce((acc, grade) => {
```
volodymyr-soltys97, line 73: You're calculating the average grade inside the sort function, which means it gets recalculated every time the sort comparison function is called. This is inefficient. It would be better to calculate the averages before sorting.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

src/sortStudents.ts
oleksandrshtonda: Can I compare here in this way?
`case SortType.Married:
        if (student1.married && !student2.married) {
          return 1 * orderNum;
        }

        if (!student1.married && student2.married) {
          return -1 * orderNum;
        }

        return 0;`

src/sortStudents.ts
oleksandrshtonda: Okay, I'm not going to do it. I have created a var copiedStudents instead :D

src/sortStudents.ts
```diff
@@ -1,16 +1,94 @@
...
+  ): number => {
+    switch (sortBy) {
+      case SortType.Name:
+        if (student1.name.toLowerCase() > student2.name.toLowerCase()) {
+          return 1 * orderNum;
+        }
+
+        if (student1.name.toLowerCase() < student2.name.toLowerCase()) {
+          return -1 * orderNum;
+        }
```
OleziO, line 50: Create a function from this code and use it.

mykhalenych (approved):

PR #1554, opened 2023-07-25
Solution

Luk2asz:


src/sortStudents.ts
```diff
@@ -1,16 +1,94 @@
...
+  const studentsCopy = [...students];
+
+  studentsCopy.sort((a: Student, b: Student) => {
+    if (sortBy === SortType.Name || sortBy === SortType.Surname) {
+      if (order === 'asc') {
+        return a[sortBy].localeCompare(b[sortBy]);
+      }
+
+      return b[sortBy].localeCompare(a[sortBy]);
+    }
```
DorotaLeniecDev, line 48: Try refactoring using switch and ternaty operator for more readability

```js
  switch (sortBy) {
    case SortType.Name:
    case SortType.Surname:
      return order === 'asc'
        ?  a[sortBy].localeCompare(b[sortBy]))
        :  b[sortBy].localeCompare(a[sortBy]));
```

DorotaLeniecDev (changes requested):
Refactor using switch

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+      case SortType.Age:
+        return order === 'asc'
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
+      case SortType.Married:
```
DorotaLeniecDev, line 49: You could treat boolean value as a number

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  });
+
+  return gradeSum / x.length;
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+
```
darokrk, line 38: you can add check variable here to simplify code:

`const sortOrder = order === 'desc' ? -1 : 1;`

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+      case SortType.Age:
+        return order === 'asc'
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
+      case SortType.Married:
```
darokrk, line 50: maybe this check can be simplify?
looks so complicated:

```
 case SortType.Married:
        if (a.married === b.married) {
          return 0;
        }

        return sortOrder * (a.married ? 1 : -1);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function averageGrade(x: number[]): number {
```
darokrk, line 22: think you can refactor this function to use reduce() method:

```
function averageGrade(x:number[]):number {
  return x.reduce((prev, cur) => {
    return prev + cur;
  }, 0) / x.length;
}
```

darokrk (changes requested):
not bad, but code needs some improvements 🛠️ 👨🏽‍💻

DorotaLeniecDev (approved):
Good job 👌

PR #1558, opened 2023-07-27
result

WhiteAshDc:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
 
 export interface Student {
   // describe Student interface
```
OleziO, line 3: Remove all comments from the code.

src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
+  };
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Age:
+      copy.sort((a, b) => (order === 'asc' ? a.age - b.age : b.age - a.age));
+      break;
+    case SortType.Name:
+      copy.sort((a, b) => (order === 'asc'
```
OleziO, line 35: Move `order === 'asc'` to the variable.

OleziO (changes requested):

VitaliyBondarenko1982 (approved):

PR #1560, opened 2023-07-29
add task solution

zvir91:


src/sortStudents.ts
```diff
@@ -1,16 +1,122 @@
...
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(students: Student[], sortBy: SortType,
+  order: SortOrder): Student[] {
+  const copySt = [...students];
+
+  const sortByName = (a: Student, b: Student) : number => {
```
Anastasiia-Svintsova, line 24: You shouldn't create a separate callback for each case. Your logic is overcomplicated a bit. For example, you don't need to repeat the same logic for the name and surname. You should check the order as well

src/sortStudents.ts
```diff
@@ -1,16 +1,122 @@
...
+    const nameA = a.name.toUpperCase();
+    const nameB = b.name.toUpperCase();
+
+    if (nameA < nameB) {
```
Anastasiia-Svintsova, line 28: You should use [.localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) for sorting string

src/sortStudents.ts
```diff
@@ -1,16 +1,122 @@
...
+  };
+
+  const sortByAverageAge = (a: Student, b: Student): number => {
+    const averageGradeA = a.grades.reduce(
+      (sum, grade) => sum + grade,
+    ) / a.grades.length;
+
+    const averageGradeB = b.grades.reduce(
+      (sum, grade) => (sum + grade),
+    ) / b.grades.length;
```
Anastasiia-Svintsova, line 79: Don't repeat the same logic twice. Create a separate method to calculate averageGrade

src/sortStudents.ts
```diff
@@ -1,16 +1,122 @@
...
+
+    case SortType.Surname:
+      return copySt.sort(sortBySurname);
+
+    case SortType.Age:
+      return copySt.sort((a: Student, b: Student): number => b.age - a.age);
+
+    case SortType.Married:
+      result1 = copySt.filter((p) => p.married === true);
+      result2 = copySt.filter((p) => p.married === false);
```
Anastasiia-Svintsova, line 110: You should sort an array, but not filter it

src/sortStudents.ts
```diff
@@ -1,16 +1,122 @@
...
+    if (order === 'asc') {
+      num = averageGradeA - averageGradeB;
+    }
+
+    return num;
+  };
+
+  let result1;
+  let result2;
+  let finalResult;
```
Anastasiia-Svintsova, line 96: Why do you need these variables? Seems it's redundant here

Anastasiia-Svintsova (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+
+      if (order === 'desc') {
+        return valueB.localeCompare(valueA);
+      }
+
+      return 0;
+    };
+  };
+
+  const sortByMarrige = (a: Student, b: Student) : number => {
```
lerastarynets, line 41: The function name 'sortByMarrige' has a spelling error. It should be 'sortByMarriage'.

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+    }
+
+    if (!marriedA && marriedB) {
+      return 1;
+    }
+
+    return 0;
+  };
+
+  const calculateAverageAge = (student: Student): number => {
```
lerastarynets, line 60: The function name 'calculateAverageAge' is misleading as it calculates the average grade, not age.

src/sortStudents.ts
```diff
@@ -1,16 +1,105 @@
...
+    return student.grades.reduce(
+      (sum, grade) => sum + grade,
+    ) / student.grades.length;
+  };
+
+  const sortByAverageAge = (a: Student, b: Student): number => {
```
lerastarynets, line 66: The function name 'sortByAverageAge' is misleading as it sorts by average grade, not age.

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #1566, opened 2023-08-07
Solution

KarolinaWiktoria:


src/sortStudents.ts
```diff
@@ -1,16 +1,103 @@
...
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(
+  students: Student[],
+  sortBy: string,
+  order: string,
+): Student[] {
+  const sortedStudents: Student[] = [...students];
```
Radoslaw-Czerniawski, line 24: Generally there is no need for explicitly typing everything everywhere, typescript is smart so if you typed the input for this function `students: Student[]` typescript will infer on it's own what type the `sortedStudents` are
```ts
const sortedStudents = [...students];
```

src/sortStudents.ts
```diff
@@ -1,16 +1,103 @@
...
+
+  switch (sortBy) {
+    case SortType.Name:
```
Radoslaw-Czerniawski, line 27: We may remove some code repetitions by declaring same operation for two cases, and using `sortBy` as accessor for the object property:
image
Take notice that I didn't type the s1 and s2, there is no need for it, as sortedStudents are already infered, typescript will know what type s1 and s2 are

src/sortStudents.ts
```diff
@@ -1,16 +1,103 @@
...
+          student1: Student,
+          student2: Student,
+        ) => (
+          order === 'asc'
+            ? student1.surname.localeCompare(student2.surname)
+            : student2.surname.localeCompare(student1.surname)
+        ),
+      );
+
+    case SortType.Age:
```
Radoslaw-Czerniawski, line 51: Same may be done for the Age and Married

src/sortStudents.ts
```diff
@@ -1,16 +1,103 @@
...
+          order === 'asc'
+            ? (+student1.married - (+student2.married))
+            : (+student2.married - (+student1.married))
+        ),
+      );
+
+    case SortType.AverageGrade:
+      return sortedStudents.sort(
+        (student1: Student, student2: Student) => {
+          const average1 = (student1.grades.reduce(
```
Radoslaw-Czerniawski, line 78: Generally this logic could be extracted to a separate function outside `sortStudents`, for example you could create a function `getAverageGrade`

image

and then use it here to simplify and reduce repetition:

image

src/sortStudents.ts
```diff
@@ -1,16 +1,103 @@
...
+          if (order === 'asc') {
+            return average1 - average2;
+          }
+
+          return average2 - average1;
+        },
+      );
 
+    default:
+      break;
```
Radoslaw-Czerniawski, line 99: Here you could add
```ts
```

Radoslaw-Czerniawski (changes requested):
Some small refactor needed :)

src/sortStudents.ts
```diff
@@ -1,16 +1,71 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
-// create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+const getAverageGrade = (
+  grades: Student['grades'],
```
darokrk, line 20: think you can use desctruction parameters here to be more clear, like:

`const getAverageGrade = ({ grades }: Student): number =>`

darokrk (changes requested):
Almost done 💪🏽  few mentions from my site 🛠️  😃

darokrk (approved):
Good job 👍🏽

PR #1567, opened 2023-08-09
Solution

HaniaNassalska:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      : studentsCopy.sort((a, b) => b.age - a.age);
+  }
+
+  if (sortBy === SortType.Married) {
+    return ascOrDesc
+      ? studentsCopy.sort((a, b) => Number(a.married) - Number(b.married))
+      : studentsCopy.sort((a, b) => Number(b.married) - Number(a.married));
+  }
+
+  const avrageGrades
```
choeqq, line 53: There's a typo in 'avrageGrades'. It should be 'averageGrades'

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+    = (arr: number[]): number => arr
```
choeqq, line 54: The 'averageGrades' function is defined inside the 'sortStudents' function. It would be better to move it outside of the 'sortStudents' function to make it reusable and keep the 'sortStudents' function cleaner.

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+  if (sortBy === SortType.Age) {
+    return ascOrDesc
+      ? studentsCopy.sort((a, b) => a.age - b.age)
+      : studentsCopy.sort((a, b) => b.age - a.age);
+  }
+
+  if (sortBy === SortType.Married) {
+    return ascOrDesc
+      ? studentsCopy.sort((a, b) => Number(a.married) - Number(b.married))
```
choeqq, line 49: Converting booleans to numbers to compare them is not best practice and may lead to confusion. Instead, you can compare directly with booleans.

src/sortStudents.ts
choeqq, line 54: The 'averageGrades' function does not handle the case of an empty array, which would result in a division by zero error. Add a check to handle this case.

choeqq (changes requested):

src/sortStudents.ts
HaniaNassalska: I really don't know how to do it better... Could you give me some advice?

choeqq (approved):
Looks good to me 🚀

PR #1568, opened 2023-08-10
Solution

SashaViun:


anastasiiavorobiova (changes requested):
Great work! Consider adding sorting by other fields: 
Name
Surname
Age
Married
You may also use the switch for this

volodymyr-soltys97 (approved):
Good job 👍

PR #1569, opened 2023-08-11
solution

Sancz0pansa:


src/sortStudents.ts
```diff
@@ -1,16 +1,109 @@
...
+
+          if (aAvg === bAvg) {
+            return 0;
+          }
+
+          if (aAvg > bAvg) {
+            return 1;
+          }
+
+          return -1;
```
asakevych, line 83: wouldn't it be enough to return aAvg - bAvg ?

asakevych (changes requested):
Good job!! but you can make some changes to make your solution smaller and easier to read 😉

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+          const bAvg = avgGrade(b);
+
+          return bAvg - aAvg;
+        });
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return undefined;
+  }
```
choeqq, line 73: I think you can create separate function for picking values and then only pass them once to sort method

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+          const aAvg = avgGrade(a);
+          const bAvg = avgGrade(b);
+
+          return bAvg - aAvg;
+        });
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return undefined;
```
choeqq, line 72: I think Error handler would be more proper here

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  AverageGrade = 'averageGrade'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function avgGrade(student: Student): number {
+  return student.grades.reduce(
+    (prev: number, item: number) => prev + item, 0,
```
choeqq, line 23: reduce is smart enough to infer types from array which uses reduce method 
```suggestion
    (prev, item) => prev + item, 0,
```

choeqq (changes requested):
Logic looks fine, but I think you can do little better - without having to type .sort method 6 times 😄

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  }
+}
+
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsCopy = [...students];
+  const orderDirection = order === 'asc';
```
mvjl000, line 47: I would make a change here. The name of this const is `orderDirection`, which suggest that it is one of two possible values. However it is boolean and its value is true only when the order is ascending.

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+      return orderDirection
+        ? studentValueA.localeCompare(studentValueB)
+        : studentValueB.localeCompare(studentValueA);
+    }
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    return orderDirection
+      ? ((studentValueA as number) - (studentValueB as number))
+      : ((studentValueB as number) - (studentValueA as number));
```
mvjl000, line 62: By using `as number` you are telling the compiler to treat it as number no matter what type it really is. If you make some change to the getValue function, add more return type or something - it might get out of control. Try to do it in some other way 🙏

mvjl000 (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+    case 'surname':
+      return student[sortBy];
+    case 'age':
+    case 'married':
+      return Number(student[sortBy]);
+    case 'averageGrade':
+      return avgGrade(student);
+    default:
+      throw new Error('Provide sort type.');
+  }
```
choeqq, line 38: please us SortType enum for cases

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+  return studentsCopy.sort((a, b) => {
+    const studentValueA = getValue(a, sortBy);
+    const studentValueB = getValue(b, sortBy);
+
+    if (typeof studentValueA === 'string'
+    && typeof studentValueB === 'string') {
+      return ascending
+        ? studentValueA.localeCompare(studentValueB)
+        : studentValueB.localeCompare(studentValueA);
+    }
```
choeqq, line 58: that should be a part of get the value, not sorting

choeqq (changes requested):

choeqq (approved):

PR #1570, opened 2023-08-11
Solution

klaudia-klis:


asakevych (changes requested):
Great job!! 🚀 
I think your solution can be even better if you use switch and group your functions: 
```
case SortType.Name:
case SortType.Surname:
   return (order === 'asc')
      ? copyOfStudents.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
      : copyOfStudents.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```
you can also create some function like `calculateAvarage` not to repeat same logic twice

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+
+    default:
+      return (order === 'asc')
+        ? copyOfStudents.sort(
+          (a, b) => calculateAverage(a.grades) - calculateAverage(b.grades),
+        )
+        : copyOfStudents.sort(
+          (a, b) => calculateAverage(b.grades) - calculateAverage(a.grades),
+        );
+  }
```
choeqq, line 61: try to use sort method only once, you can have separate function for picking values to sort, only then pass them to sort method

choeqq (changes requested):
Logic looks really good, I think you could just clean up your code, with function for picking values, instead of using sort method 10 times 🚀

mvjl000 (approved):
Well done! 🚀

PR #1572, opened 2023-08-13
Develop

Horizon-git:


src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+  Married = 'married',
+  AverageGrade = 'grades',
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
+
+function getAvarageGrade(grades: number[]): number {
+  return grades.reduce((acc, grade) => acc + grade) / grades.length;
```
volodymyr-soltys97, line 21: In the getAvarageGrade function, you're using reduce without an initial value. This is risky as it may lead to an error when the grades array is empty. It's safer to set the initial value to zero.

src/sortStudents.ts
```diff
@@ -1,16 +1,56 @@
...
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  return [...students].sort((student1: Student, student2: Student): number => {
+    let firstToCompare: Student = student1;
+    let secondToCompare: Student = student2;
+
+    if (order === 'desc') {
```
volodymyr-soltys97, line 33: You can move this magic word to enum and use it everywher

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #1573, opened 2023-08-14
add task solution

ogospadzipalina:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+    return orderType
+      ? studentsCopy.sort((a, b) => a.name.localeCompare(b.name))
+      : studentsCopy.sort((a, b) => b.name.localeCompare(a.name));
+  }
+
+  if (sortBy === SortType.Surname) {
+    return orderType
+      ? studentsCopy.sort((a, b) => a.surname.localeCompare(b.surname))
+      : studentsCopy.sort((a, b) => b.surname.localeCompare(a.surname));
+  }
```
asakevych, line 47: 1. please use [switch case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) for it 
2.  your cases can be grouped - 2 numeric, 2 string and 1 avarage 

```suggestion
case SortType.Name:
case SortType.Surname:
   return (order === 'asc')
      ? studentsCopy.sort((a, b) => a[sortBy].localeCompare(b[sortBy]))
      : studentsCopy.sort((a, b) => b[sortBy].localeCompare(a[sortBy]));
```

asakevych (changes requested):
Good job! just some small changes needed 😉

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+    case 'average':
+      return orderType
+        ? studentsCopy
+          .sort((a, b) => AverageGrade(a.grades) - AverageGrade(b.grades))
+        : studentsCopy
+          .sort((a, b) => AverageGrade(b.grades) - AverageGrade(a.grades));
+
+    default:
+      throw new Error('no such sort type');
+  }
```
choeqq, line 59: please try to think of a way to use .sort method once - I would create function for picking values, and then pass them only once into .sort method

choeqq (changes requested):
Looking good, but I think you can figure out the way to use sort method only once 🚀

choeqq (approved):

PR #1574, opened 2023-08-14
i get this weird error but code looks good to me, message in the erro…

szymonpachucki:
…r file  @mate-academy/students-ts-config not found

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,55 @@
-
 export interface Student {
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
 }
+
+const averageGrades = (arrOfGrades: number[]): number => {
+  if (arrOfGrades.length === 0) {
+    return 0;
+  }
+
+  return arrOfGrades.reduce((sum: number, grade: number) => sum + grade, 0) / arrOfGrades.length;
+};
 
 export enum SortType {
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
+      }
 
+      return copyOfStudent.sort((a, b) => Number(b.married)
+        - Number(a.married));
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+    default:
+      return students;
+  }
```
choeqq, line 75: try to create a separate function for picking values, then pass them to sort method, like that you won't have to use sort 10 times 😄

src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
+export type SortOrder = 'asc' | 'desc';
+
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
+  const copyOfStudent = [...students];
+  const orderType = order === 'asc';
+
+  switch (sortBy) {
+    case SortType.Name:
+      if (orderType) {
```
choeqq, line 35: you can just reverse sorting order depending on orderType instead of checking it in each case

choeqq (changes requested):

choeqq (changes requested):

choeqq (approved):
Looks good to me ✅

PR #1580, opened 2023-08-14
Develop

Matylda-Lewandowska:
Task Solution - sort students

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,61 @@
 
 export interface Student {
-  // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
 }
 
 export enum SortType {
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+    if (a[sortBy] > b[sortBy]) {
+      return order === 'asc' ? 1 : -1;
+    }
+
+    if (a[sortBy] < b[sortBy]) {
+      return order === 'asc' ? -1 : 1;
+    }
+
+    return 0;
+  });
```
choeqq, line 58: I think switch would be useful here

choeqq (approved):

src/sortStudents.ts
Matylda-Lewandowska: I will change it :)

PR #1581, opened 2023-08-14
Develop

Meg-Sowka:


src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+
+    case SortType.AverageGrade:
+      studentsArr.sort((a, b) => {
+        return isAsc
+          ? getAvarageGrade(a.grades) - getAvarageGrade(b.grades)
+          : getAvarageGrade(b.grades) - getAvarageGrade(a.grades);
+      });
+      break;
+
+    default:
```
choeqq, line 73: please provide some deafault return / error handler

src/sortStudents.ts
```diff
@@ -1,16 +1,78 @@
...
+      break;
+  }
```
choeqq, line 75: try to create separate function for picking value, then just pass picked value to sort method once

choeqq (changes requested):
Looks good but I think you can have a lot less code, and bettter readibility

mvjl000 (changes requested):
Looks good to me! 🚀
 I would just make one small change...

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const studentsArr = [...students];
+  const isAsc: boolean = order === 'asc';
+
+  function compareFunction(a: Student, b: Student): number {
+    // const aMarried = Number(a.married);
+    // const bMarried = Number(b.married);
+
```
choeqq, line 37: remove unecessary comments

choeqq (changes requested):

choeqq (approved):
Well done 🚀

PR #1586, opened 2023-08-15
sort students

dimkamg21:


src/sortStudents.ts
```diff
@@ -1,16 +1,59 @@
...
-  // write your function
+  return sum / student.grades.length;
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): object[] {
+  const studentsCopy = [...students];
+
+  const sortingFunction = (a: Student, b: Student): number => {
```
Viktor-Kost, line 32: let's oprtimize your solution. 
1. no declare sort method like this
```
  return studentsCopy.sort((a, b) => (
    order === 'asc' ? sortingFunction(a, b) : sortingFunction(b, a)
  ));
```
2. If the solutions are the same for different cases you can use a group
3. No needs to create one more variable  such as `studentsCopy ` you can return at once

Example: 
```
export function sortStudents(
  students : Student[],
  sortBy : SortType,
  order : SortOrder,
) : Student[] {
  return [...students].sort((a, b) => {
    switch (sortBy) {
      case SortType.Name:
      case SortType.Surname:
        return (
          order === 'asc'
            : b[sortBy].localeCompare(a[sortBy])
        );
        
        ........
        {{continue for other cases}}
       
    }
  });
  ```

Viktor-Kost (changes requested):

maxim2310 (approved):
Well done!!!

PR #1588, opened 2023-08-16
solution

patryk177m:


src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+
+    case SortType.AverageGrade:
+      return sortByAsc
+        ? copyStudents.sort((student1, student2) =>
+          avgGrades(student1.grades) - avgGrades(student2.grades))
+        : copyStudents.sort((student1, student2) =>
+          avgGrades(student2.grades) - avgGrades(student1.grades));
+    default:
+      return copyStudents;
+  }
```
choeqq, line 71: I would suggest creating separate function for picking VALUES to sort, then pass them to native `.sort` method, with sorting direction

src/sortStudents.ts
```diff
@@ -1,16 +1,72 @@
...
+  Married,
+  AverageGrade,
 }
 
-// create SortOrder type
-export type SortOrder;
+const avgGrades = (v: Array<number>): number => {
+  return v.length === 0
+    ? 0
+    : v.reduce((acc: number, grade: number) => acc + grade, 0) / v.length;
```
choeqq, line 22:
```suggestion
    : v.reduce((acc grade) => acc + grade, 0) / v.length;
```
reduce is smart enough to infer types from passed array, you don't have to type them manually 😉

choeqq (changes requested):
Looks good but I think you can make it little more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) 🚀

choeqq (changes requested):
Waiting for any updates @patryk177m 👀

choeqq (approved):
Looks good to me ✅

PR #1592, opened 2023-08-16
sortStudents solution

anpiet:


src/sortStudents.ts
```diff
@@ -1,16 +1,70 @@
...
+    // sort by AverageGrade
+    case (sortBy === 'grades' && order === 'asc'):
+      return newArr.sort((a, b) => (a.grades.reduce((c, d) => c + d, 0)
+      / a.grades.length) - ((b.grades.reduce((c, d) => c + d, 0))
+      / b.grades.length));
 
+    case (sortBy === 'grades' && order === 'desc'):
+      return newArr.sort((a, b) => (b.grades.reduce((c, d) => c + d, 0)
+      / b.grades.length) - ((a.grades.reduce((c, d) => c + d, 0))
+      / a.grades.length));
```
choeqq, line 65: please try to figure out the way not to repeat that much code 

1. your cases are correct, maybe use them just to get the value instead of sorting isntantly
2. get the value from comparisons like: 

     - +!!b.married - +!!a.married
     - b.age - a.age
     - a.surname.localeCompare(b.surname)
3. Pass the comparison result to sort method
4. change the sorting direction depending on sortOrder

choeqq (changes requested):

src/sortStudents.ts
anpiet: Thanks for comments. I think that now code is much more simple and clear. Please let me know if you think that something else should be changed here.

choeqq (approved):
Looks good to me ✅

PR #1595, opened 2023-08-16
add task solution

MateuszSlezak:


src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+
+    case SortType.Married:
+      return order === 'asc'
+        ? studCopy.sort((a, b) => (a.married ? -1 : 1) - (b.married ? -1 : 1))
+        : studCopy.sort((a, b) => (a.married ? -1 : 1) - (b.married ? -1 : 1));
 
+    case SortType.AverageGrade:
+      return order === 'asc'
+        ? studCopy.sort((a, b) => calcGrade(a.grades) - calcGrade(b.grades))
+        : studCopy.sort((a, b) => calcGrade(b.grades) - calcGrade(a.grades));
```
choeqq, line 63: please try to think of less repetetive way to sort those values, for example: 

1. sort method needs only comparison result of 2 values
2. create function to compare those values 
3. pass the comparison function result to sort
4. change sorting direction depending on `order`

choeqq (changes requested):

choeqq (approved):
Looks good to me ✅

PR #1597, opened 2023-08-17
solution

MaciejKujawa:


src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+    return order === 'asc'
+      ? studentsCopy.sort((a, b) => Number(a.married) - Number(b.married))
+      : studentsCopy.sort((a, b) => Number(b.married) - Number(a.married));
+  }
+
+  if (sortBy === SortType.AverageGrade) {
+    return order === 'asc'
+      ? studentsCopy.sort((a, b) => averageGrade(a) - averageGrade(b))
+      : studentsCopy.sort((a, b) => averageGrade(b) - averageGrade(a));
+  }
```
choeqq, line 61: try to think of better way to implelement sorting - you're repeating a lot of sort logic, maybe extract only values, return from comparison, like those ones:

- Number(a.married) - Number(b.married))
- b.surname.localeCompare(a.surname));
- averageGrade(a) - averageGrade(b)

And only then use sort, depending on those values

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updares @MaciejKujawa 👀

choeqq (changes requested):
Waiting for any updates @MaciejKujawa 👀

PR #1598, opened 2023-08-18
solution

Capoverde:
I hope it works :)


src/sortStudents.ts
```diff
@@ -1,16 +1,63 @@
...
+        const averageGradeA = averageGrade(a.grades);
+        const averageGradeB = averageGrade(b.grades);
+
+        return averageGradeA - averageGradeB;
+      }
+      default:
+
+        return 0;
+    }
+  };
```
asakevych, line 56: 1. some cases has empty line before some not - make it everywhere the same
2. you can group your cases - for example Name and Surname are strings so you can use same callback - `a[sortBy].localeCompare(b[sortBy]);`

asakevych (changes requested):
Great job! just small changes needed 😉 remember about 'clean code', it's important code to be easy to read

choeqq (approved):
Looks good to me ✅

PR #1599, opened 2023-08-19
Add solution

ghost:


src/sortStudent.test.ts
choeqq, line 1: why are you changing test file? 🤔

choeqq (approved):

PR #1602, opened 2023-09-02
Implemented sorting of students

davinnchii:


src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder,
+): Student[] {
+  const result: Student[] = [...students];
+
+  return result.sort((a: Student, b: Student) => {
+    if (sortBy === SortType.AverageGrade) {
```
BudnikOleksii, line 32: As we have different sort types let's use switch case

src/sortStudents.ts
```diff
@@ -1,16 +1,62 @@
...
+      return order === 'desc'
+        ? -1
+        : 1;
+    }
 
+    if (a[sortBy] < b[sortBy]) {
+      return order === 'desc'
+        ? 1
+        : -1;
+    }
```
BudnikOleksii, line 58: It's difficult to understand in what cases we compare like this. Also, if you compare strings use `localeCompare` instead.

BudnikOleksii (changes requested):
Let's get rid of so many `if` statements, and use a switch case instead, it fits perfectly with enums

PR #1604, opened 2023-09-18
Solution

Vasyl-Zhyliakov:


src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+        if (a.married === b.married) {
+          return 0;
+        }
+
+        if (a.married) {
+          return 1;
+        }
+
+        return -1;
+      };
```
VitaliyBodnar, line 51: can be less 
```suggestion
      comparator = (a, b): number => Number(a.married) - Number(b.married)
```

src/sortStudents.ts
```diff
@@ -1,16 +1,69 @@
...
+      break;
+    case SortType.AverageGrade:
+      comparator = (a, b): number => {
+        return getAverageGrade(a.grades) - getAverageGrade(b.grades);
+      };
+      break;
+    default:
+      return [];
```
VitaliyBodnar, line 59: you need to return the original list in a default case

VitaliyBodnar (changes requested):

SanyaBratashchuk (approved):

PR #1606, opened 2023-09-22
Develop

YuraMisura:


src/sortStudents.ts
```diff
@@ -1,16 +1,51 @@
...
+      case SortType.Name:
+      case SortType.Surname:
+        return a[sortBy].localeCompare(b[sortBy]);
+
+      case SortType.Age:
+        return a[sortBy] - b[sortBy];
+
+      case SortType.Married:
+        return a[sortBy] && !b[sortBy] ? -1 : 1;
+
```
etojeDenys, line 39:
```suggestion

      case SortType.Age:
      case SortType.Married:
        return +a[sortBy] - +b[sortBy];
```
we can group these cases also

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #1611, opened 2023-10-15
ts sort students - first try

RufatRM:


anastasiia-tilikina (changes requested):
Good job with typing.

But your code contains too many duplicates though.
Basically, you have two types of sorting: as strings, and as numbers.
You can use switch case right inside sort method (instead of if-else)
and if you have desc order just swap params (hint: you can use [destructuring assignment](https://javascript.info/destructuring-assignment#array-destructuring))

Example:
```typescript
copiedStudents.sort((a, b) => {
  if (order === 'desc') {
    // swap params;
}

  switch (sortType) {
    case SortTupe.Name:
      // continue
  }
})
```

src/sortStudents.ts
```diff
@@ -1,16 +1,64 @@
...
+        return studentA.married
+          .toString().localeCompare(studentB.married.toString());
+
+      case SortType.AverageGrade:
+        return averageA - averageB;
 
+      default:
+        return 0;
+    }
+  });
```
anastasiia-tilikina, line 61: you forgot to use a, b vars instead of studentA and studentB.

and combine all cases where you use the same sorting (as string)

anastasiia-tilikina (changes requested):
Almost done!

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1612, opened 2023-10-30
ts_sort_students

netzelnatalia:


src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+
+export function sortStudents(
+  students: Array<Student>,
+  sortBy: SortType,
+  order: SortOrder,
+): Array<Student> {
+  const studentsCopy = [...students];
+
+  return studentsCopy.sort((a: Student, b: Student): number => {
+    const direction = order === 'desc' ? -1 : 1;
```
anastasiiavorobiova, line 34: Consider splitting a logic for sort and direction.
For example, you can apply sort and go through the switch block without paying attention to the direction. After the switch, you can create an additional if block, that checks a direction and applies the reverse array method if needed.

anastasiiavorobiova (changes requested):
Great work! Let's make it perfect!

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+
+    switch (sortBy) {
+      case SortType.Name:
+        return a.name.localeCompare(b.name);
```
anastasiiavorobiova: This case always sorts in one direction

src/sortStudents.ts
```diff
@@ -1,16 +1,61 @@
...
+      case SortType.Surname:
+        return a.surname.localeCompare(b.surname);
```
anastasiiavorobiova: This case always sorts in one direction

anastasiiavorobiova (changes requested):
Look carefully through the review comments, and feel free to ask for help in the chat

anastasiiavorobiova (approved):
Great work!

PR #1613, opened 2023-11-06
Solution

ArtemVelychko:


anastasiia-tilikina (changes requested):
Great job!
Still we have things to improve)

1. move switch case inside sort method to avoid code duplication
2. use enum values as cases

```typescript
studentsCopy.sort((a, b) => {
  switch (sortBy) {
    case SortType.Name:
      return a.name.localeCompare(b.name)
// and so on
})
``` 

3. here is a hint how to avoid ternary operators. we can use [destructuring assignment](https://javascript.info/destructuring-assignment#array-destructuring) to swap vars:

```typescript
studentsCopy.sort((a, b) => {
  let [studentA, studentB] = [a, b]

  if (order === 'desc') {
    [studentA, studentB] = [studentB, studentA]
  }

  switch // ... your code
 })
```

anastasiia-tilikina (approved):
Well done!

PR #1615, opened 2023-11-08
Solution

alina-kabanets:


anastasiia-tilikina (approved):
Great job!

Also, have a look at this implementation as well. We can use [destructuring assignment](https://javascript.info/destructuring-assignment) to swap variables

```typescript
let [studentA, studentB] = [a,b];

if (order === 'desc') {
  [studentA, studentB] = [studentB, studentA];
}

``` 
Your solution is correct as well) just consider that there is one more way for the future)

PR #1620, opened 2023-11-23
Solution

zaura333:


src/sortStudents.ts
```diff
@@ -1,16 +1,75 @@
...
+export function sortStudents(
+  students: Array<Student>,
+  sortBy: SortType,
+  order: SortOrder,
+): Array<Student> {
+  let sortingCallback;
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (sortBy === SortType.Name || sortBy === SortType.Surname) {
```
Barosz30: There's a a checklist that says to use switch.
You can use 2 cases if they got same body like this:
Case SortType.Name:
Case SortType.Surname:
  (Propably some sorting here)

Barosz30 (approved):
It works and wasnt realy easy task, so i approve. Tough check checklist

mbruhler (approved):
Well done, approved

PR #1621, opened 2023-11-23
solution

Barosz30:
So.... Much... Sorting....

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -1,16 +1,65 @@
 
 export interface Student {
   // describe Student interface
+  name: string,
+  surname: string,
+  age: number,
+  married: boolean,
+  grades: number[],
 }
 
 export enum SortType {
   // describe SortType enum
+  Name = 'name',
+  Surname = 'surname',
+  Age = 'age',
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+// for use, in case of AverageGrade sorting
+export function Avg(student: Student): number {
+  return student.grades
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+export function Avg(student: Student): number {
+  return student.grades
+    .reduce((sum, grade) => sum + grade, 0) / student.grades.length;
+}
 
-export function sortStudents(students, sortBy, order) {
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder) :Student[] {
   // write your function
+  let newarray: number;
```
DanyloSemenov: The code looks good!
The variable `newarray` is suggested to be renamed to something like `compareResult` to make its purpose clearer.

DanyloSemenov (changes requested):

DanyloSemenov (approved):

src/sortStudents.ts
mbruhler: yeah, true

mbruhler (changes requested):
Change the variable names to more human understandable and it will be approved

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+  Married = 'married',
+  AverageGrade = 'grades'
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+// for use, in case of AverageGrade sorting
+export function Avg(student: Student): number {
```
JoyCoffeeAddict: the only thing I would change is to rename this function to 
`getAverageGrade` this way it's self explanatory and you don't even need the comment.

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
This looks good!

PR #1623, opened 2023-11-23
solution

bertsu02:


adrian200002 (approved):
It took me a while to understand how sortOrderFactor works in line 72, interesting solution. The code works, so for me, it's fine. Nice work.

src/sortStudents.ts
```diff
@@ -1,16 +1,76 @@
...
+
+      default:
+        throw new Error('Invalid SortType');
+    }
+  }
+
+  sortedStudents.sort((a: Student, b: Student) => {
+    const sortOrderFactor = order === 'asc' ? 1 : -1;
 
+    return sortOrderFactor * sorter(a, b);
```
mbruhler, line 72: well done here, clean solution

mbruhler (approved):
GJ Approved.

One this is that we can arithmetically calculate boolean values like `false - true === -1`. That would reduce amount of code, but you used a function here which for me is acceptable.

PR #1624, opened 2023-11-23
Solution

Sibunas:


szymanski-marcin (approved):
It works! but one thing i'd change is var - try to replace it with const/let instead

Sibunas:
> It works! but one thing i'd change is var - try to replace it with const/let instead

Do You mean aVar and bVar? It's only name of Variables (a, b and 3 letters from Variable).

szymanski-marcin:
watched at the wrong file when i put that comment. Too much for today, my bad🤣

Sibunas:
> watched at the wrong file when i put that comment. Too much for today, my bad🤣

Oh ok no problem I fully understand You :D

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+        aVar = a.surname;
+        bVar = b.surname;
+        break;
+      case SortType.Age:
+        aVar = a.age;
+        bVar = b.age;
+        break;
+      case SortType.Married:
+        aVar = a.married;
+        bVar = b.married;
```
mvjl000: Could you try this?

```suggestion
      case SortType.Name:
      case SortType.Surname:
      case SortType.Age:
      case SortType.Married:
        aVar = a[sortBy];
        bVar = b[sortBy];
        break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+        break;
+      case SortType.AverageGrade:
+        aVar = a.grades
+          .reduce((acc, grade) => acc + grade, 0) / a.grades.length;
+
+        bVar = b.grades
+          .reduce((acc, grade) => acc + grade, 0) / b.grades.length;
```
mvjl000: Repeating same code. Let's create a function that calculates the average for us!

mvjl000 (changes requested):
That's an interesting approach. I have some improvements for your solution

src/sortStudents.ts
Sibunas: That's one nice, didn't know it can by done like that.

src/sortStudents.ts
Sibunas: That one is applied too.

mvjl000 (approved):
Good!

mvjl000 (approved):

mpiskadlo (approved):

PR #1625, opened 2023-11-23
Solution

patrykyo20:
The solution failed one test but when I did it manually it was seemingly all right

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -31,27 +31,23 @@ export function sortStudents(
     switch (sortBy) {
       case SortType.Name:
         result = order === 'asc'
-          ? a.name.localeCompare(b.name)
-          : b.name.localeCompare(a.name);
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
         break;
       case SortType.Surname:
         result = order === 'asc'
-          ? a.surname.localeCompare(b.surname)
-          : b.surname.localeCompare(a.surname);
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
         break;
       case SortType.Age:
         result = order === 'asc'
-          ? a.age - b.age
-          : b.age - a.age;
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
         break;
       case SortType.Married:
-        if (a.married && !b.married) {
-          result = order === 'asc' ? -1 : 1;
-        } else if (!a.married && b.married) {
-          result = order === 'asc' ? 1 : -1;
-        } else {
-          result = 0;
-        }
+        result = order === 'asc'
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
         break;
       case SortType.AverageGrade:
         result = order === 'asc'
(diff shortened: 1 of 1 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+      case SortType.Age:
+        result = order === 'asc'
+          ? a.age - b.age
+          : b.age - a.age;
+        break;
+      case SortType.Married:
+        if (a.married && !b.married) {
+          result = order === 'asc' ? -1 : 1;
+        } else if (!a.married && b.married) {
+          result = order === 'asc' ? 1 : -1;
```
zaura333: Try switching minus signs, I think right now expected results for asc and desc are switched

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+      case SortType.Name:
+        result = order === 'asc'
+          ? a.name.localeCompare(b.name)
+          : b.name.localeCompare(a.name);
+        break;
+      case SortType.Surname:
+        result = order === 'asc'
+          ? a.surname.localeCompare(b.surname)
+          : b.surname.localeCompare(a.surname);
+        break;
```
zaura333: There you could use fall-through since code for both cases is the same, just then it'd have to be a[sortBy] instead of a.name and a.surname

zaura333 (changes requested):
Overall looks good! I'm requesting changes so that you can check sorting order in that one case

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  return [...students].sort((a: Student, b: Student) => {
+    let result = 0;
+    const aAvg: number = a.grades.reduce(
+      (acc: number, prev: number) => acc + prev, 0,
+    ) / a.grades.length;
+    const bAvg: number = b.grades.reduce(
+      (acc: number, prev: number) => acc + prev, 0,
+    ) / b.grades.length;
```
mvjl000: Please notice that `aAvg` and `bAvg` is calculated even if `sortBy` is not `grades`. IMO it makes sense to create a separate function for it and calculate it inside of `case` (or simply move it there)

mvjl000 (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,65 @@
...
+        result = order === 'asc'
+          ? a[sortBy].localeCompare(b[sortBy])
+          : b[sortBy].localeCompare(a[sortBy]);
+        break;
+      case SortType.Age:
+        result = order === 'asc'
+          ? a[sortBy] - b[sortBy]
+          : b[sortBy] - a[sortBy];
+        break;
+      case SortType.Married:
```
mbruhler, line 49: This raises and TS error that left-side operator must be number-like type.

In order to fix, cast boolean to number
```
    case SortType.Married:
        result = order === 'asc'
          : Number(b[sortBy]) - Number(a[sortBy])
```

in plain JS it would work but here we are, in type-safety universe :)
It's a clever solution that you wanted to subtract boolean values, gj

mbruhler (changes requested):
One minor change needed

mvjl000 (approved):
Well done!

PR #1628, opened 2023-11-23
solution

MarcinLigmann:


jacekkosiba (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,101 @@
...
 
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
```
mbruhler: This can be one switch statement 
```
switch() {
    case SortType.Name:
    case SortType.Surname:
         //code
         break;
}
```

src/sortStudents.ts
```diff
@@ -1,16 +1,101 @@
...
+          return b.age - a.age;
+        });
+      break;
+
+    case SortType.Married:
+      sortedStudents
+        .sort((a: Student, b: Student) => {
+          if (order === 'asc') {
+            if (a.married < b.married) {
+              return -1;
```
mbruhler: In overall, to determine ascending and descending you can just get the calculation of a field and multiply it by 1 or -1

This applies everywhere and will reduce the amount of code

mbruhler (changes requested):
Your code has some issues:
- there is too much ascending descending handling. Use 1 and -1 and multiply the field calculation result to get the proper ordering
- group logically similar switch statements into one as I show in comments

JoyCoffeeAddict (approved):

PR #1630, opened 2023-11-23
add task solution

kate-prakofyeva:


Timi2424 (approved):

mbruhler (changes requested):
Ordering can be just 1 if asc or -1 if descending, this should be a multiplier passed to the function and used for calculations so you get the proper ordering. This will reduce amount of code here. Create a function for calculations and pass it to the sort function as callback
 
```
      return order === 'asc'
        ? copyArray.sort((a, b) => +a[sortBy] - +b[sortBy])
        : copyArray.sort((a, b) => +b[sortBy] - +a[sortBy]);
```

kate-prakofyeva:
I have a question:
what to do in this situation:
image
I just call it order2

mbruhler (approved):
Now it's cool, approved

kate-prakofyeva:
> I have a question: what to do in this situation: image I just call it order2

@mbruhler to jest prawidlowy zapis?

PR #1631, opened 2023-11-23
solution

jacekkosiba:


bertsu02 (approved):
You love those switch cases don't you 😄 
All clean, good job 🚀

src/sortStudents.ts
```diff
@@ -1,16 +1,73 @@
...
+        result = a.name.localeCompare(b.name);
+        break;
+      case SortType.Surname:
+        result = a.surname.localeCompare(b.surname);
+        break;
+      case SortType.Age:
+        result = a.age - b.age;
+        break;
+      case SortType.Married:
+        if (a.married === b.married) {
```
mbruhler: you can also calculate this arithmetically. `false - true === -1`, `true - false === 1`

mbruhler (changes requested):
One minor change and we're good to go!

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+      case SortType.AverageGrade:
+        avgGradeA = a.grades.reduce((acc, curr) => {
+          return acc + curr;
+        }) / a.grades.length;
+
+        avgGradeB = b.grades.reduce((acc, curr) => {
+          return acc + curr;
+        }) / b.grades.length;
+
+        result = avgGradeA - avgGradeB;
```
JoyCoffeeAddict: make it a separate function so it's more readable, and then you will not need additional variables `avgGradeA` at the top of a compare function. 
These vars concern only the case `SortType.AverageGrade` so they should not be available for the rest of the function. You can declare variables in cases, just add `{}` to a `case`

JoyCoffeeAddict (changes requested):

src/sortStudents.ts
jacekkosiba: Great tip, thanks!

JoyCoffeeAddict (approved):

PR #1632, opened 2023-11-23
Solution

RadoslawBielski:


src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+
+    switch (sortBy) {
+      case SortType.Name:
+        comparison = a.name.localeCompare(b.name);
+        break;
+      case SortType.Surname:
+        comparison = a.surname.localeCompare(b.surname);
+        break;
+      case SortType.Age:
+        comparison = a.age - b.age;
```
patrykyo20: You can change all instances of dot notation, such as a.name and a.surname, to bracket notation like a[sortBy] instead.

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+        break;
+      case SortType.Married:
+        if (a.married === b.married) {
+          comparison = 0;
+        } else if (a.married) {
+          comparison = 1;
+        } else {
+          comparison = -1;
+        }
```
patrykyo20: In this case, it's only necessary to use` a[sortBy] - b[sortBy].` You can simplify the comparison using the ternary operator, like this: `comparison = a.married === b.married ? 0 : a.married ? 1 : -1`

patrykyo20 (approved):
Slight changes, but the task works perfectly.

mvjl000 (changes requested):
Good job Radek but still your solution can be brought to another level! Try to implement changes suggested by Patryk

src/sortStudents.ts
```diff
@@ -1,16 +1,74 @@
...
+  grades: number[];
 }
 
 export enum SortType {
-  // describe SortType enum
+  Name,
+  Surname,
+  Age,
+  Married,
+  AverageGrade,
```
mvjl000: Add values to this enum cause by default those are gonna be numbers from 1 to N
```suggestion
  Name = 'name',
... etc
```

mvjl000 (changes requested):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+        break;
+      case SortType.Age:
+        comparison = a.age - b.age;
+        break;
+      case SortType.Married:
+        comparison = Number(a.married) - Number(b.married);
+        break;
+      case SortType.AverageGrade:
+        avgGradeA = calculateAverage(a.grades);
+        avgGradeB = calculateAverage(b.grades);
```
mvjl000: You don't need to declare those variables at the top of the function. Simply do it here ;)

mvjl000 (changes requested):
One small tweak ;)

mbruhler (approved):
Code works well, and is well written too!
Accepted!

PR #1635, opened 2023-11-26
Solution

marcinzrodlowski:


mosowiecka (approved):
If you have predetermined amount of options it is better to use ```switch/case``` structure instead
of multiple ```if/else``` blocks

JoyCoffeeAddict (approved):

PR #1637, opened 2023-11-27
solution

aureliagorzna:


michal-reszka (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+  sortedStudents.sort((a, b) => {
+    switch (sortBy) {
+      case SortType.Name:
+        return order === 'asc'
+          ? a.name.localeCompare(b.name)
+          : b.name.localeCompare(a.name);
+      case SortType.Surname:
+        return order === 'asc'
+          ? a.surname.localeCompare(b.surname)
+          : b.surname.localeCompare(a.surname);
```
mvjl000: Those two cases could be combined into one
```suggestion
      case SortType.Name:
      case SortType.Surname:
        return order === 'asc'
          ? a[sortBy].localeCompare(b[sortBy])
          : b[sortBy].localeCompare(a[sortBy]);
```

src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
+}
+
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
+
+  sortedStudents.sort((a, b) => {
+    switch (sortBy) {
+      case SortType.Name:
```
mvjl000, line 32: Note that in each `case` you are checking whether the order is ascending or descending. This makes a lot of additional lines of code, which could be avoided. Could you find a different way to do it?

mvjl000 (changes requested):

src/sortStudents.ts
mbruhler: That's still an issue. Maybe multiply by -1 and 1 depending on sorting order and for example 
`order * a[sortBy].localeCompare(b[sortBy])`, something like that

mbruhler (changes requested):
Please check my comment, and suggestion, how can u improve the sorting order

mpiskadlo (approved):

PR #1642, opened 2023-12-11
First version

spencerspe:


src/sortStudents.ts
```diff
@@ -1,16 +1,91 @@
...
   // write your function
+  const newStude = students.map((studier) => studier);
+  let result: Student[] = [];
+
+  function average(student: Student): number {
+    return student.grades.reduce((acc: number, val: number) => (acc + val),
+      0) / student.grades.length;
+  }
+
+  if (sortBy === 'grades') {
```
lerastarynets: pls move all sortBy properties to an enum and use it's values where needed

src/sortStudents.ts
```diff
@@ -1,16 +1,91 @@
...
+    result = newStude.sort((a, b) => {
+      if (order === 'asc') {
+        return average(a) - average(b);
+      }
+
+      return average(b) - average(a);
+    });
```
lerastarynets: pls consider using ternaries
```suggestion
  order === 'asc' ? average(a) - average(b) : average(b) - average(a);
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
-
-export function sortStudents(students, sortBy, order) {
+export function sortStudents(students: Student[],
+  sortBy: SortType, order: SortOrder): Student[] {
   // write your function
+  const newStude = students.map((studier) => studier);
```
anastasiia-tilikina: why do you need this map?

src/sortStudents.ts
```diff
@@ -1,16 +1,81 @@
...
+  let result: Student[] = [];
```
anastasiia-tilikina, line 25: create here the copy of the original array and mutate it.
```suggestion
  let result: Student[] = [...students];
```

anastasiia-tilikina (changes requested):
good job!
but let's improve your code

we have too many if statements, therefore it is hard to read.
lets replace them with switch case structure, and use it inside sort method to avoid code duplication. Also we can use [destructuring assignment](https://javascript.info/destructuring-assignment#array-destructuring) and check order only once:

```typescript
result.sort((a, b) =>{
  let [studenaA, studentB] = [a, b];

  if (order === 'desc') {
    [studenaA, studentB] = [b, a]; // we don't need to check order again
  }

  switch(sortBy) {
    case SortType.Name: 
    // ...
  }
})
``` 

with this approach our code will be cleaner, shorter and easier to read.

lerastarynets (approved):
lgtm 🚀

PR #1647, opened 2024-01-21
Add task solution

Laisens:


src/sortStudents.ts
```diff
@@ -1,16 +1,140 @@
...
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  switch (sortBy) {
+    case SortType.Name:
+      copyStudents.sort((a, b) => {
+        const valueA = a[SortType.Name];
+        const valueB = b[SortType.Name];
+
+        if (order === 'asc') {
```
anastasiiavorobiova: Consider moving repeating logic into a separate variable (and assigning 1 or -1) you can multiply the sorting result and this variable in the switch case

anastasiiavorobiova (changes requested):
Great work! Almost done!

volodymyr-soltys97 (approved):
Great work 🔥

PR #1649, opened 2024-01-25
mySolution

AntonFof:


src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+  return [...students].sort((a, b) => {
+    let comparison = 0;
+
+    switch (sortBy) {
+      case SortType.Name:
+        comparison = a.name.localeCompare(b.name);
+        break;
+      case SortType.Surname:
+        comparison = a.surname.localeCompare(b.surname);
+        break;
```
etojeDenys: you can group these cases
```suggestion
      case SortType.Name:
      case SortType.Surname:
        comparison = a[sortBy].localeCompare(b[sortBy]);
        break;
```

src/sortStudents.ts
```diff
@@ -1,16 +1,60 @@
...
+      case SortType.Age:
+        comparison = a.age - b.age;
+        break;
+      case SortType.Married:
+        if (a.married && !b.married) {
+          comparison = 1;
+        } else if (!a.married && b.married) {
+          comparison = -1;
+        }
+        break;
```
etojeDenys: same here

etojeDenys: think how to group these cases

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
GJ!

AntonFof:
Thanks!

PR #1656, opened 2024-02-28
add task solution

pawelnowicki87:


src/sortStudents.ts
```diff
@@ -1,16 +1,67 @@
...
 
+export function sortStudents(
+  students: Student[], sortBy: SortType, order: SortOrder,
+): Student[] {
+  const sortedStudents = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  sortedStudents.sort((a, b) => {
+    let comparison = 0;
```
yevhenii-pyl, line 28: Interesting approach with that `comparison` variable. But do you actually need it? Doesn't it just complicates things? 
Basically, you can simplify it by just returning from the switch case, like:

```
    case SortType.Name:
      sortedStudents = [...students].sort((a: Student, b: Student) => {
        if (order === 'asc') {
          return a.name.localeCompare(b.name);
        }

        return b.name.localeCompare(a.name);
      });
      break;
```

However, this approach forces you to consider asc | desc order. 

Thus, I'll just leave it to you to decide what works better for you. 
Anyways, good solid job here 👍

yevhenii-pyl (approved):

PR #1658, opened 2024-03-12
Created a sortStudents

YarynaPuhach:


TideTwo (approved):
You could probably use a ternary op to decide between `asc` and `desc` order. You use both `type[]` and `Array<type>` for arrays, would look more consistent if you stuck to 1, and you could use `toSorted` if you don't want to modify the original array. (these are semantic suggestions for next time, the task is reviewed as approved (for peer review anyway))

choeqq (approved):

PR #1660, opened 2024-03-12
Solution

Davwidos:


TideTwo (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+  Age,
+  Married,
+  AverageGrade
 }
 
 // create SortOrder type
-export type SortOrder;
+export type SortOrder = 'asc' | 'desc';
 
+function compareStrings(a: string, b: string, order: SortOrder): number {
```
TideTwo, line 23: Could have access to order without the argument, if declared inside `sortStudents` (if you wanted to do so).  You could also use `toSorted` to avoid mutating original array.

choeqq (approved):

PR #1663, opened 2024-03-13
Sort Students

itsXenthis:


src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
+  Married = 'married',
+  AverageGrade = 'averageGrade',
+}
+
+export type SortOrder = 'asc' | 'desc';
+
+function averageGrade(student: Student): number {
+  return student.grades.reduce((sum, currentStudentGrade) => (
+    sum + currentStudentGrade
+  ), 0) / student.grades.length;
```
Davwidos, line 22: It's good practice to check if `student.grades.length` is not equal 0 before you divide.

Davwidos (approved):

src/sortStudents.ts
```diff
@@ -1,16 +1,66 @@
...
-// create SortOrder type
-export type SortOrder;
+export function sortStudents(
+  students: Student[],
+  sortBy: SortType,
+  order: SortOrder = 'asc',
+): Student[] {
+  let sortedStudentArr = [...students];
+
+  sortedStudentArr = sortedStudentArr.sort((
```
nataliaklonowska, line 32: Method sort mutates the array, so you don't need to reassign the result of the sorting

nataliaklonowska (changes requested):

src/sortStudents.ts
nataliaklonowska: Method sort mutates the array, so you don't need to reassign the result of the sorting

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #1668, opened 2024-03-13
first version

konradwasilewski444:


marinatea (approved):
all seems okay;)

nataliaklonowska (changes requested):
If you have predetermined amount of options it is better to use switch/case structure instead of multiple if/else blocks

choeqq (approved):

PR #1673, opened 2024-04-12
Solution

aMtiIV:


src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
 
 export interface Student {
   // describe Student interface
```
BogdanMaliuta: remove comments

src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
+
+  switch (order) {
+    case 'asc':
+      sortOrder = 1;
+      break;
+    case 'desc':
+      sortOrder = -1;
+      break;
+    default:
+      return [];
```
BogdanMaliuta: why empty array?

src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
+      break;
+    case SortType.Surname:
+      sortCallback = (
+        firstStudent: Student,
+        secondStudent: Student,
+      ): number => firstStudent
+        .surname
+        .localeCompare(secondStudent.surname)
+        * sortOrder;
+      break;
```
BogdanMaliuta: you can do this using only 1 case like 
```suggestion
case SortType.Name:
case SortType.Surname:
```

think how to make it and, do this also with next cases

src/sortStudents.ts
```diff
@@ -1,16 +1,96 @@
...
+          .reduce((sum: number, grade: number) => sum + grade, 0)
+        / firstStudent.grades.length
+        - secondStudent
+          .grades
+          .reduce((sum: number, grade: number) => sum + grade, 0)
+        / secondStudent.grades.length
+      ) * sortOrder;
+      break;
+    default:
+      return [];
```
BogdanMaliuta: why again empty arr?

BogdanMaliuta (changes requested):
Hello, a lot of works was done.But code is very hard to read, needs to be completely refactor, removing reduntant variables like sortOrder, and make things described in my comments

BogdanMaliuta (approved):
Nice!

PR #1675, opened 2024-04-20
task solution

keex1k:


src/sortStudents.ts
```diff
@@ -1,16 +1,140 @@
...
-  // describe Student interface
+  name: string;
+  surname: string;
+  age: number;
+  married: boolean;
+  grades: number[];
 }
 
 export enum SortType {
   // describe SortType enum
```
BogdanMaliuta: remove comments

src/sortStudents.ts
```diff
@@ -1,16 +1,140 @@
...
+          if (nazwaA < nazwaB) {
+            return -1;
+          }
+
+          if (nazwaA > nazwaB) {
+            return 1;
+          }
+
+          return 0; // Jeśli są równe
+        });
```
BogdanMaliuta: You have same logic in this 2 cases, you can combine this, also for sorting straings use localeCompare method

src/sortStudents.ts
```diff
@@ -1,16 +1,140 @@
...
+          if (avgA > avgB) {
+            return -1;
+          }
+
+          if (avgA < avgB) {
+            return 1;
+          }
+
+          return 0; // Jeśli są równe
+        });
```
BogdanMaliuta: huge mistake to use if else here, its code duplicating, by you need to use default order

BogdanMaliuta (changes requested):
Hello, not bad, but needs to be refactored

keex1k:
> Hello, not bad, but needs to be refactored

Hi, I refactored my code, please check it

src/sortStudents.ts
```diff
@@ -1,16 +1,82 @@
...
+export function sortStudents(students: Student[],
+  sortBy: SortType,
+  order: SortOrder): Object[] {
+  const result = [...students];
 
-export function sortStudents(students, sortBy, order) {
-  // write your function
+  if (order === 'asc') {
+    switch (sortBy) {
+      case SortType.Name:
```
yevhenii-pyl: You can combine SortType.Name && SortType.Surname at one condition: 

```
case SortType.Name:
case SortType.Surname:
     result.sort((a, b) => a[sortBy].localeCompare(b[sortBy]));
```

&& we should check the `order` here. We really want to remove that 'if-else' you have. So the final solution for these cases may look like this: 

```
      result.sort((a, b) => (
        order === 'asc'
```

yevhenii-pyl (changes requested):
We have to remove if-else from your switch statement. 

Please, check the example with name && surename sorting and adjust this logic to other cases.

yevhenii-pyl (approved):
You got it! 🥇

PR #1680, opened 2024-07-01
My solution to the task

karaniewskiruslan:
Created a `sortStudents` function that takes 3 parameters and return a new array
with students sorted according to the given params. The original array must not
change.

### DIFF
--- a/src/sortStudents.ts
+++ b/src/sortStudents.ts
@@ -7,11 +7,11 @@ export interface Student {
 }
 
 export enum SortType {
-  Name = 'name',
-  Surname = 'surname',
-  Age = 'age',
-  Married = 'married',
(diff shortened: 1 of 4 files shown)


src/sortStudents.ts
```diff
@@ -1,16 +1,68 @@
...
+          comparison = 0;
+          break;
+        }
+
+        if (a.married) {
+          comparison = 1;
+          break;
+        }
+
+        comparison = -1;
```
zibi-dev, line 54: this can be simply put like that also because `Number(true) === 1` and `Number(false) === 0`
```suggestion
       comparison = Number(a.married) - Number(b.married);
```

Zibi95 (approved):
Well done!

